PR #1, opened 2022-07-12
Fix compiler warnings about read API

gagachang:
1. Check the return value when semu calls read API to read content
from STDIN. If there is an error or EOF, we can ignore further
UART operations.

2. Check the return value when semu calls fread API to read
content from binary or image. If the size of read content is less
than requested size, we trigger a fatal error to exit semu.

### DIFF
--- a/semu.c
+++ b/semu.c
@@ -305,7 +305,8 @@ static void *uart_thread_func(void *priv)
     struct uart *uart = (struct uart *) priv;
     while (1) {
         char c;
-        read(STDIN_FILENO, &c, 1);
+        if (read(STDIN_FILENO, &c, 1) <= 0)  /* an error or EOF */
+            continue;
         pthread_mutex_lock(&uart->lock);
         while ((uart->data[UART_LSR - UART_BASE] & UART_LSR_RX) == 1)
             pthread_cond_wait(&uart->cond, &uart->lock);
@@ -1303,7 +1304,9 @@ size_t read_file(FILE *f, uint8_t **r)
     fseek(f, 0, SEEK_SET);
 
     uint8_t *content = malloc(fsize + 1);
-    fread(content, fsize, 1, f);
+    if (fread(content, fsize, 1, f) != 1) /* less than fsize bytes */
+        fatal("read file content");
+
     content[fsize] = 0;
     *r = content;
 


semu.c
```diff
@@ -305,7 +305,9 @@ static void *uart_thread_func(void *priv)
     struct uart *uart = (struct uart *) priv;
     while (1) {
         char c;
-        read(STDIN_FILENO, &c, 1);
+        ssize_t bytes_read = read(STDIN_FILENO, &c, 1);
+        if (bytes_read == -1 || bytes_read == 0)  // An error or EOF
```
jserv, line 309: Use C-style comment. i.e., `/* this is a comment */`.

jserv: How about the following change?
```diff
@@ -305,7 +305,8 @@ static void *uart_thread_func(void *priv)
     struct uart *uart = (struct uart *) priv;
     while (1) {
         char c;
-        read(STDIN_FILENO, &c, 1);
+        if (read(STDIN_FILENO, &c, 1) <= 0)
+            continue;
         pthread_mutex_lock(&uart->lock);
         while ((uart->data[UART_LSR - UART_BASE] & UART_LSR_RX) == 1)
             pthread_cond_wait(&uart->cond, &uart->lock);
```

semu.c
```diff
@@ -1303,7 +1305,10 @@ size_t read_file(FILE *f, uint8_t **r)
     fseek(f, 0, SEEK_SET);
 
     uint8_t *content = malloc(fsize + 1);
-    fread(content, fsize, 1, f);
+    size_t read_blocks = fread(content, fsize, 1, f);
```
jserv, line 1308: Ditto. I would expect the following:
```diff
@@ -1303,7 +1304,8 @@ size_t read_file(FILE *f, uint8_t **r)
     fseek(f, 0, SEEK_SET);
 
     uint8_t *content = malloc(fsize + 1);
-    fread(content, fsize, 1, f);
+    if (fread(content, fsize, 1, f) != 1) /* less than fsize bytes */
+        fatal("read file content");
     content[fsize] = 0;
     *r = content;
 
```

jserv (changes requested):
The patch not only eliminated compilation warnings but also fixes the usage of read(2). You should clarify in the subject of this pull request and git commit message.

semu.c
gagachang: > How about the following change?
> 
> ```diff
> @@ -305,7 +305,8 @@ static void *uart_thread_func(void *priv)
>      struct uart *uart = (struct uart *) priv;
>      while (1) {
>          char c;
> -        read(STDIN_FILENO, &c, 1);
> +        if (read(STDIN_FILENO, &c, 1) <= 0)
> +            continue;
>          pthread_mutex_lock(&uart->lock);
>          while ((uart->data[UART_LSR - UART_BASE] & UART_LSR_RX) == 1)
>              pthread_cond_wait(&uart->cond, &uart->lock);
> ```

Yes I agree. It's more concise.

semu.c
gagachang: I have applied it in new commit

jserv:

PR #3, opened 2022-07-15
Integrate riscv-tests

gagachang:
As the discussion in the issue ticket.
This patch first supports **./semu  --test** command to run all test binaries and show the test result like:

```
$ make test
cc -O2 -Wall -Iinclude -o semu semu.c test/test.c -lpthread
./semu --test
Test binary rv64ui_p_fence_i failed...
An exception occurred.
Test binary rv64ua_p_amoadd_w failed...
Fail test case = 2.
Test binary rv64ua_p_amoand_d failed...
An exception occurred.
Pass: rv64ui_p_add
Pass: rv64ui_p_addi
Pass: rv64ui_p_addiw
Pass: rv64ui_p_addw
Pass: rv64ui_p_and
Pass: rv64ui_p_andi
Pass: rv64ui_p_auipc
Fail: rv64ui_p_fence_i
...
Pass: rv64ui_p_jal
...
Pass: rv64ua_p_amoadd_d
Fail: rv64ua_p_amoadd_w
Fail: rv64ua_p_amoand_d
=======================
Test result: 51/54
=======================
```

A new entry point is declared to execute test flow.
And a new C file **./test/test.c** and related header are created for information of test binaries.
Note that I just put parts of the test binaires into this patch.

I think there are many spaces to optimize this patch.
Please review this patch and give some advices, thanks.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/riscv-tests"]
+	path = tests/riscv-tests
+	url = https://github.com/riscv-software-src/riscv-tests
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,34 @@ CC ?= gcc
 CFLAGS = -O2 -Wall
 LDFLAGS = -lpthread
 
-all:
-	$(CC) $(CFLAGS) -o semu semu.c $(LDFLAGS)
+# For compiling riscv-tests
+CROSS_COMPILE ?= riscv64-unknown-elf-
+
+CUR_DIR := $(shell pwd)
+TESTS_DIR := $(CUR_DIR)/tests
+RISCV_TESTS_DIR := $(TESTS_DIR)/riscv-tests
+RISCV_TESTS_TARGET_DIR := $(RISCV_TESTS_DIR)/target
+RISCV_TESTS_SUITE_DIR := $(RISCV_TESTS_TARGET_DIR)/share/riscv-tests/isa
+RISCV_TESTS_BIN_DIR := $(TESTS_DIR)/riscv-tests-data
+
+BIN := semu
+
+OBJS := semu.o
+
+# Whether to enable riscv-tests
+ENABLE_RISCV_TESTS ?= 0
+ifeq ("$(ENABLE_RISCV_TESTS)", "1")
+CFLAGS += -DENABLE_RISCV_TESTS
+OBJS += tests/isa-test.o
+endif
+
+%.o: %.c
+	$(CC) -o $@ $(CFLAGS) -c $<
+
+all: $(BIN)
+
+$(BIN): $(OBJS)
+	$(CC) -o $@ $^ $(LDFLAGS)
 
 kernel.bin:
 	scripts/download.sh
@@ -11,7 +37,31 @@ kernel.bin:
 check: all kernel.bin
 	./semu kernel.bin fs.img
 
+tests/riscv-tests/configure:
+	git submodule update --init --recursive
+
+# Fetch and build riscv-tests project
+# Transform the original elf format to binary format
+build-riscv-tests: tests/riscv-tests/configure
+	cd $(RISCV_TESTS_DIR); autoconf; ./configure --prefix=$(RISCV_TESTS_TARGET_DIR)
+	$(MAKE) -C $(RISCV_TESTS_DIR)
+	$(MAKE) -C $(RISCV_TESTS_DIR) install
+	mkdir -p $(RISCV_TESTS_BIN_DIR)
+	for file in $(RISCV_TESTS_SUITE_DIR)/rv64*; do \
+		case $$file in \
+			(*.dump) continue; \
+		esac; \
+		original=$$(basename $$file); \
+		filename=$$(echo $$original | sed -e "s/-/_/g"); \
+		$(CROSS_COMPILE)objcopy -O binary $$file $(RISCV_TESTS_BIN_DIR)/$$filename; \
+		echo $$file "--> $(RISCV_TESTS_BIN_DIR)/$$filename"; \
+	done
+
+riscv-tests: $(BIN) build-riscv-tests
+	./semu --test
+
 clean:
-	rm -f semu
+	rm -f $(BIN) $(OBJS)
 distclean: clean
 	rm -f kernel.bin fs.img
+	rm -rf $(RISCV_TESTS_BIN_DIR)
--- a/semu.c
+++ b/semu.c
@@ -6,6 +6,10 @@
 #include <string.h>
 #include <unistd.h>
 
+#ifdef ENABLE_RISCV_TESTS
+#include "tests/test.h"
+#endif
+
 /* xv6 uses only 128MiB of memory. */
 #define RAM_SIZE (1024 * 1024 * 128)
 
@@ -120,6 +124,14 @@ struct ram *ram_new(const uint8_t *code, const size_t code_size)
     memcpy(ram->data, code, code_size);
     return ram;
 }
+
+#ifdef ENABLE_RISCV_TESTS
+void ram_free(struct ram *ram)
+{
+    free(ram->data);
+    free(ram);
+}
+#endif
 
 exception_t ram_load(const struct ram *ram,
                      const uint64_t addr,
@@ -621,9 +633,12 @@ void bus_disk_access(struct bus *bus)
 
 typedef enum { USER = 0x0, SUPERVISOR = 0x1, MACHINE = 0x3 } cpu_mode_t;
 
+#define NUM_REG 32
+#define NUM_CSR 4096
+
 struct cpu {
-    uint64_t regs[32], pc;
-    uint64_t csrs[4096];
+    uint64_t regs[NUM_REG], pc;
+    uint64_t csrs[NUM_CSR];
     cpu_mode_t mode;
     struct bus *bus;
     bool enable_paging;
@@ -1313,13 +1328,102 @@ size_t read_file(FILE *f, uint8_t **r)
     return fsize;
 }
 
-int main(int argc, char **argv)
+#ifdef ENABLE_RISCV_TESTS
+#define TOHOST_ADDR 0x80001000
+#define TOHOST_EXCEPTION_MAGIC 1337
+
+int semu_test_start(int argc, char **argv)
 {
-    if (argc < 2) {
-        printf("Usage: %s <filename> [<image>]\n", argv[0]);
-        return 2;
+    struct cpu *cpu = NULL;
+    uint64_t tohost = 0;
+
+    /* iterate test data */
+    for (int n = 0; n < n_riscv_tests; n++) {
+        FILE *f = fopen(riscv_tests[n].file_path, "rb");
+        if (!f)
+            fatal("open test binary file");
+
+        uint8_t *binary = NULL;
+        size_t fsize = read_file(f, &binary);
+        fclose(f);
+
+        if (!cpu) {
+            cpu = cpu_new(binary, fsize,
+                          NULL); /* allocate new CPU for first time */
+        } else {
+            /* Reset CPU and free RAM */
+            ram_free(cpu->bus->ram);
+            memset(cpu->regs, 0, NUM_REG * sizeof(cpu->regs[0]));
+            memset(cpu->csrs, 0, NUM_CSR * sizeof(cpu->csrs[0]));
+            cpu->regs[2] = RAM_BASE + RAM_SIZE;
+            cpu->pc = RAM_BASE, cpu->mode = MACHINE;
+            /* New RAM content */
+            cpu->bus->ram = ram_new(binary, fsize);
+        }
+        free(binary);
+
+        while (1) {
+            /*
+             * Once the test is done, the test binary writes a value into
+             * tohost variable, which is located at address 0x80001000.
+             *
+             * When each test stars, tohost will be initialized as 0.
+             * If tohost is changed to 1, it menas SEMU passes all the test.
+             * If tohost is less than 1337, the fail test number will be
+             * (tohost >> 1)
+             * If tohost is greater than 1337, there are some exceptions
+             * occurred.
+             */
+            if (bus_load(cpu->bus, TOHOST_ADDR, 64, &tohost) != OK)
+                break;
+            if (tohost != 0)
+                break;
+
+            /* Fetch instruction */
+            uint64_t insn;
+            exception_t e;
+            if ((e = cpu_fetch(cpu, &insn)) != OK) {
+                cpu_take_trap(cpu, e, NONE);
+                if (exception_is_fatal(e))
+                    break;
+                insn = 0;
+            }
+
+            cpu->pc += 4; /* advance pc */
+
+            /* decode and execute */
+            if ((e = cpu_execute(cpu, insn)) != OK) {
+                cpu_take_trap(cpu, e, NONE);
+                if (exception_is_fatal(e))
+                    break;
+            }
+        }
+
+        /*
+         * Test result is stored at a0 (x10).
+         * The riscv-tests set a0 to 0 when all tests pass.
+         * Otherwise it indicates a failure or exception.
+         */
+        if (cpu->regs[10] == 0) {
+            riscv_tests[n].result = TEST_PASS;
+        } else {
+            printf("\nTest binary %s fail...\n", riscv_tests[n].name);
+            printf("a0 = 0x%lx\n", cpu->regs[10]);
+            printf("tohost = 0x%lx\n", tohost);
+            if (tohost < TOHOST_EXCEPTION_MAGIC)
+                printf("Fail test case = %ld.\n", tohost >> 1);
+            else
+                printf("An exception occurred.\n");
+        }
     }
 
+    print_test_result();
+    return 0;
+}
+#endif
+
+int semu_start(int argc, char **argv)
+{
     FILE *f = fopen(argv[1], "rb");
     if (!f)
         fatal("open kernel image file");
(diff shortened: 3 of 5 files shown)


scripts/build_riscv_tests.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+OBJCOPY=${RISCV}/bin/riscv64-unknown-elf-objcopy
```
jserv, line 3: Introduce the variable `CROSS_COMPILE`.

Makefile
```diff
@@ -1,16 +1,22 @@
...
+	$(CC) $(CFLAGS) -o semu semu.c test/test.c $(LDFLAGS)
 
 kernel.bin:
 	scripts/download.sh
 
 check: all kernel.bin
 	./semu kernel.bin fs.img
 
+build_test_data:
+	scripts/build_riscv_tests.sh
```
jserv, line 15: Check file permission. Executable.

Makefile
```diff
@@ -1,16 +1,22 @@
...
+
+test: all
```
jserv, line 17: The directory name should be `tests`, and the target in `Makefile` would be `riscv-tests`.

Makefile
jserv, line 15: Add proper `git clone` to get `riscv-tests` repository here.

semu.c
```diff
@@ -1313,13 +1324,98 @@ size_t read_file(FILE *f, uint8_t **r)
...
+                    break;
+            }
+        }
+
+        /*
+         * Test result is stored at a0 (x10).
+         * The riscv-tests set a0 to 0 when all tests pass.
+         * Otherwise it indicates a failure or exception.
+         */
+        // printf("a0 = 0x%lx\n", cpu->regs[10]);
```
jserv, line 1401: Don't leave the dead code.

test/test.c
```diff
@@ -0,0 +1,95 @@
+#include <stdio.h>
+
+#include "test.h"
+
+struct testdata riscv_tests[] = {
+    add_testdata("rv64ui_p_add", "./test/riscv-tests-data/rv64ui_p_add"),
```
jserv, line 6: Replace all occurrences of `./test/` with `tests/`.

jserv (changes requested):
Follow the directory structure of rv32emu.
See https://github.com/sysprog21/rv32emu/tree/master/tests

Avoid preserving the binary blob in git repository. Instead, generate via script(s).

Makefile
```diff
@@ -1,16 +1,22 @@
 CC ?= gcc
-CFLAGS = -O2 -Wall
+CFLAGS = -O2 -Wall -Iinclude
 LDFLAGS = -lpthread
 
 all:
```
jserv, line 5: Modularize the build system. i.e., there should be something like `ENABLE_RISCV_TESTS` which can be turned on/off on demand.
See https://github.com/sysprog21/rv32emu/blob/master/Makefile

semu.c
```diff
@@ -1313,13 +1324,98 @@ size_t read_file(FILE *f, uint8_t **r)
     return fsize;
 }
 
-int main(int argc, char **argv)
+void semu_test_start(int argc, char **argv)
```
jserv, line 1327: Add the statement before the test oriented functions. i.e.,
```c
#ifdef ENABLE_RISCV_TESTS
void semu_test_start(int argc, char **argv)
{
```

semu.c
```diff
@@ -1313,13 +1324,98 @@ size_t read_file(FILE *f, uint8_t **r)
...
+             * If tohost is greater than 1337, there are some exceptions
+             * occurred.
+             */
+            if (bus_load(cpu->bus, 0x80001000, 64, &tohost) != OK)
+                break;
+            else if (tohost != 0)
+                break;
+
+            /* Fetch instruction */
+            uint64_t inst;
```
jserv, line 1377: Use `insn` instead of `inst`. See `git log` for details.

include/test.h
```diff
@@ -0,0 +1,27 @@
+#ifndef __TEST_H__
```
jserv, line 1: Move `include/test.h` to `tests/test.h` and don't rely on `-I` option. Specify the relative path -- `#include "tests/test.h"`

gagachang:
Thanks, I need some time to modify them.
Please wait for my update.

include/test.h
gagachang: done

semu.c
gagachang: done

scripts/build_riscv_tests.sh
gagachang: Done, CROSS_COMPILE is defined in Makefile

Makefile
gagachang: I removed this script file since Makefile is enough for us

gagachang: The riscv-tests is submodule now.

semu.c
gagachang: Fixed.

test/test.c
gagachang: Done

Makefile
gagachang: ENABLE_RISCV_TESTS is added into Makefile now.
The test.o is only compiled and linked into semu when this flag equals one.

Makefile
```diff
@@ -2,16 +2,63 @@ CC ?= gcc
 CFLAGS = -O2 -Wall
 LDFLAGS = -lpthread
 
-all:
-	$(CC) $(CFLAGS) -o semu semu.c $(LDFLAGS)
+# For compiling riscv-tests
+CROSS_COMPILE = riscv64-unknown-elf-
```
jserv, line 6: Change to `CROSS_COMPILE ?= riscv64-unknown-elf-` which allows overrides.

tests/test.c
```diff
@@ -0,0 +1,125 @@
+#include <stdio.h>
```
jserv, line 1: Rename `tests/test.c` to `tests/isa-test.c` which emphasizes on its purpose.

semu.c
```diff
@@ -1363,6 +1468,24 @@ int main(int argc, char **argv)
...
+    if (argc < 2) {
+        printf("Usage: %s <filename> [<image>]\n", argv[0]);
+#ifdef ENABLE_RISCV_TESTS
+        printf("Usage: %s --test\n", argv[0]);
+#endif
+        return 2;
+    }
+
+#ifdef ENABLE_RISCV_TESTS
+    if (!strncmp(argv[1], "--test", strlen("--test")))
```
jserv, line 1484: You can simply hardcode `6` instead of invoking `strlen("--test")` since it is not necessary to make function calls here.

semu.c
```diff
@@ -1363,6 +1468,24 @@ int main(int argc, char **argv)
...
+        semu_test_start(argc, argv);
```
jserv, line 1485: Once you change the function prototype to `int semu_test_start(int argc, char **argv)`, you can rewrite as following:
```c
    if (!strncmp(argv[1], "--test", 6))
        return semu_test_start(argc, argv);

```

semu.c
```diff
@@ -1363,6 +1468,24 @@ int main(int argc, char **argv)
...
+    else
+#endif
+        semu_start(argc, argv);
```
jserv, line 1488: Ditto. If you change the function prototype of semu_start, you can simplify as following:
```c
    return semu_start(argc, argv);
```

Makefile
```diff
@@ -2,16 +2,63 @@ CC ?= gcc
...
 
 kernel.bin:
 	scripts/download.sh
 
 check: all kernel.bin
 	./semu kernel.bin fs.img
 
+# Fetch and build riscv-tests project
+# Transform the original elf format to binary format
+build-riscv-tests:
```
jserv, line 42: Add a new target:
```
tests/riscv-tests:
    git submodule update --init
```

And `build-riscv-tests` target should depend on tests/riscv-tests.

tests/test.h
```diff
@@ -0,0 +1,27 @@
...
+#define TEST_PASS 0
+#define TEST_FAIL -1
+
+struct testdata {
+    int result;
+    const char *name;
+    const char *file_path;
+};
+
+#define add_testdata(n, path)                              \
```
jserv, line 13: A macro should have its name all capitalize. i.e., `ADD_INSN_TEST`.
Since we assume the directory structure, we can rewrite as following:
```c
#define ADD_INSN_TEST(op) \
    { .result = TEST_FAIL, \
      .name = op, \
      .file_path = "tests/riscv-tests-data/" #op, \
    }
```

See https://gcc.gnu.org/onlinedocs/gcc-12.1.0/cpp/Stringizing.html

tests/test.c
```diff
@@ -0,0 +1,125 @@
...
+}
+
+void print_test_result(void)
+{
+    int pass = 0;
+    int testdata_n = get_testdata_n();
+
+    for (int n = 0; n < testdata_n; n++) {
+        if (riscv_tests[n].result != TEST_PASS) {
+            printf("\033[0;31mFail: %s\033[0m\n", riscv_tests[n].name);
```
jserv, line 115: Move ANSI control codes to dedicated macros.
See https://github.com/sysprog21/threaded-logger/blob/master/colors.h

tests/test.c
```diff
@@ -0,0 +1,125 @@
...
+    add_testdata("rv64ua_p_amoswap_w",
+                 "tests/riscv-tests-data/rv64ua_p_amoswap_w"),
+    add_testdata("rv64ua_p_amoxor_d",
+                 "tests/riscv-tests-data/rv64ua_p_amoxor_d"),
+    add_testdata("rv64ua_p_amoxor_w",
+                 "tests/riscv-tests-data/rv64ua_p_amoxor_w"),
+    add_testdata("rv64ua_p_lrsc", "tests/riscv-tests-data/rv64ua_p_lrsc"),
+};
+
+struct testdata *get_testdata(void)
```
jserv, line 97: It is not worth defining two functions which refer to static data. Drop them and use macro such as `ARRAY_SIZE`.

tests/test.c
```diff
@@ -0,0 +1,125 @@
+#include <stdio.h>
+
+#include "test.h"
+
+struct testdata riscv_tests[] = {
```
jserv, line 5: Add modifier `static const`.

semu.c
```diff
@@ -1313,13 +1328,103 @@ size_t read_file(FILE *f, uint8_t **r)
     return fsize;
 }
 
-int main(int argc, char **argv)
+#ifdef ENABLE_RISCV_TESTS
+#define TOHOST_ADDR 0x80001000
+#define TOHOST_EXCEPTION_MAGIC 1337
+
```
jserv, line 1334: Add `extern` here.
```c
extern struct testdata riscv_tests[];
```

gagachang: Hi, I found a problem.
If we add ```extern``` here, we can not use sizeof() to calculate the size of riscv_tests[] in compile time.
How about maintaining another variable called ```riscv_tests_nr``` to store the size and ```extern``` it?

```
int riscv_tests_nr = sizeof(riscv_tests) / sizeof(struct testdata);
extern int riscv_tests_nr;
```

tests/test.c
gagachang: ```struct testdata``` has a **result** member to store the test result, and it is initialized as TEST_FAIL.
When SEMU runs a test correctly, it will assign result as TEST_PASS.
So this array could not be const if we have this purpose.

```
struct testdata {
    int result;
    const char *name;
    const char *file_path;
};
```

Is it really necessary to be const variable?

jserv: Both field `name` and `file_path` must be declared as `const` while field `result` can be writable.

semu.c
jserv: Using another `extern` variable sounds fine. Rename it to `n_riscv_tests` for short.

Makefile
gagachang: Done!

semu.c
gagachang: Done!

semu.c
gagachang: Done!

Makefile
gagachang: Done

tests/test.c
gagachang: Done. Add three macros for reset color, dark red and dark green over this function.

semu.c
gagachang: Done. The two ```extern``` variables are defined in ```test/test.h```.

semu.c
```diff
@@ -1313,13 +1328,102 @@ size_t read_file(FILE *f, uint8_t **r)
...
+             * When each test stars, tohost will be initialized as 0.
+             * If tohost is changed to 1, it menas SEMU passes all the test.
+             * If tohost is less than 1337, the fail test number will be
+             * (tohost >> 1)
+             * If tohost is greater than 1337, there are some exceptions
+             * occurred.
+             */
+            if (bus_load(cpu->bus, TOHOST_ADDR, 64, &tohost) != OK)
+                break;
+            else if (tohost != 0)
```
jserv, line 1379: Replace "else if" with "if" for readability.

semu.c
```diff
@@ -1366,3 +1470,21 @@ int main(int argc, char **argv)
...
+#ifdef ENABLE_RISCV_TESTS
+        printf("Usage: %s --test\n", argv[0]);
+#endif
+        return 2;
+    }
+
+#ifdef ENABLE_RISCV_TESTS
+    if (!strncmp(argv[1], "--test", 6))
+        return semu_test_start(argc, argv);
+    else
```
jserv, line 1487: Remove this "else" since it is not necessary.

Makefile
```diff
@@ -2,16 +2,65 @@ CC ?= gcc
...
+		case $$file in \
+			(*.dump) continue; \
+		esac; \
+		original=$$(basename $$file); \
+		filename=$$(echo $$original | sed -e "s/-/_/g"); \
+		$(CROSS_COMPILE)objcopy -O binary $$file $(RISCV_TESTS_BIN_DIR)/$$filename; \
+		echo $$file "--> $(RISCV_TESTS_BIN_DIR)/$$filename"; \
+	done
+
+riscv-tests: all build-riscv-tests
```
jserv, line 60: Target `riscv-tests` should depend on `$(BIN)` rather than `all` because we might extend the goal of `all` later.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tests/riscv-tests"]
+	path = tests/riscv-tests
+	url = https://github.com/riscv-software-src/riscv-tests.git
```
jserv, line 3: Get rid of suffix `.git` from `url`. i.e., specify `url = https://github.com/riscv-software-src/riscv-tests`

Makefile
gagachang: OK, I think we also need to remove test binary folder when we ```make distclean```.

```
distclean: clean
	rm -f kernel.bin fs.img
	rm -rf $(RISCV_TESTS_BIN_DIR)
```

Makefile
```diff
@@ -2,16 +2,65 @@ CC ?= gcc
...
+$(BIN): $(OBJS)
+	$(CC) -o $@ $^ $(LDFLAGS)
 
 kernel.bin:
 	scripts/download.sh
 
 check: all kernel.bin
 	./semu kernel.bin fs.img
 
+tests/riscv-tests:
```
jserv, line 40: Replace `tests/riscv-tests` with `tests/riscv-tests/configure` because we have to track the files in directory `riscv-tests`.

Makefile
```diff
@@ -2,16 +2,65 @@ CC ?= gcc
...
+	git submodule update --init --recursive
+
+# Fetch and build riscv-tests project
+# Transform the original elf format to binary format
+build-riscv-tests: tests/riscv-tests
+	cd $(RISCV_TESTS_DIR); autoconf; ./configure --prefix=$(RISCV_TESTS_TARGET_DIR)
```
jserv, line 46: Is it required to run `autoconf` before the execution of `configure` script?

Makefile
jserv: Exactly, let `dist-clean` target work as expected.

tests/isa-test.c
```diff
@@ -0,0 +1,104 @@
...
+
+    for (int n = 0; n < n_riscv_tests; n++) {
+        if (riscv_tests[n].result != TEST_PASS) {
+            printf(C_DR "Fail: %s" C_RST "\n", riscv_tests[n].name);
+        } else {
+            printf(C_DG "Pass: %s" C_RST "\n", riscv_tests[n].name);
+            pass++;
+        }
+    }
+
```
jserv, line 100: Add a new blank line before displaying test results.

Makefile
gagachang: I just follow the steps provided from riscv-tests.
The README of riscv-tests says:

```
$ git clone https://github.com/riscv/riscv-tests
$ cd riscv-tests
$ git submodule update --init --recursive
$ autoconf
$ ./configure --prefix=$RISCV/target
$ make
$ make install
```

jserv (changes requested):
Add the reference test result at the end of git commit message. e.g.,
```
=======================
Test result: 53/71
=======================
```

The git HEAD of riscv-tests should be mentioned as well.

Makefile
jserv: Check if "autoconf" is necessary, otherwise we have to document automake/autoconf as dependent packages.

jserv: In addition, check if `make install` is a necessary step.

jserv: We might run riscv-tests with GitHub Actions, so that we can validate pull requests automatically. See [SSE2NEON](https://github.com/DLTcollab/sse2neon/actions) as an example.
It would be vital if we can eliminate unnecessary commands/steps since the shortened CI pipeline implies less time to wait.

jserv: Can you skip the generation of RV32 specific files?

gagachang: Sure.
Let me pull a new patch for other comments.
I will check build flow of riscv-tests later.

tests/isa-test.c
gagachang: Done. Now we display test result as:

```
...
Fail: rv64ua_p_amoxor_w
Fail: rv64ua_p_lrsc

=======================
Test result: 53/71
=======================
```

Makefile
gagachang: Done

Makefile
gagachang: > Can you skip the generation of RV32 specific files?

Refer to their [isa test Makefile](https://github.com/riscv-software-src/riscv-tests/blob/master/isa/Makefile), it seems they always build RV32 tests. Only RV64 is optional.

But we can build isa tests only  by
```
make isa
```

Also, ```make install``` is unnecessary since it just copies generated files into another directory.

PR #4, opened 2022-07-16
Make use of 'curl' in macOS

bugorz:
### Note to reviewers
 - Make download script work on macOS where `curl` is installed by default.
 - Thank you for making this emulator :)

### Tests
 - On my Mac (macOS Monterey 12.4):
```Shell
➜  semu git:(master) ✗ make distclean check
rm -f semu
rm -f kernel.bin fs.img
cc -O2 -Wall -o semu semu.c -lpthread
scripts/download.sh
Downloading...
kernel.bin: OK
fs.img: OK
./semu kernel.bin fs.img

xv6 kernel is booting

init: starting sh
$ echo
echo
$ ^Cmake: *** [check] Interrupt: 2
```

### DIFF
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -10,11 +10,7 @@ ROOTFS_CHECKSUM=27fd366b4383f1176f44095fa270d23e9b3f4deb
 
 echo "Downloading..."
 
-if [[ $OSTYPE == 'darwin'* ]]; then
-	DOWNLOAD_CMD="curl -s -O -L -C -"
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Let's use curl instead. No checking for wget is required.

bugorz:
@jserv Switch to curl as suggested, thanks.

jserv:
Thank @bugorz for contributing!

PR #5, opened 2022-07-17
Optimize riscv-tests build flow

gagachang:
Only build isa tests of riscv-tests for saving build time.

Also, autoconf is unnecessary since there is a pre-generated configure
file in riscv-tests repository. See issue #273:
https://github.com/riscv-software-src/riscv-tests/issues/273


jserv:

PR #6, opened 2022-07-18
Fix RV32A instructions

gagachang:
Refer to RISC-V specification: "The RISC-V Instruction Set Manual
Volume I: Unprivileged ISA", Section 8.4:

    For RV64, 32-bit AMOs always sign-extend the value placed in rd.

And the recent test result of riscv-tests becomes:

```
Pass: rv64ua_p_amoadd_w
Pass: rv64ua_p_amoswap_w

=======================
Test result: 55/71
=======================
```


jserv:

PR #7, opened 2022-07-21
Add option to test specific test binary

gagachang:
Current test command always tests whole test cases. This patch
adds an option to run just one specific test case. It is useful
when we plan to re-run certain instruction test(s). For example:

    ./semu --test rv64ui_p_add

Also, this patch makes the log of test result more pretty.
Now the test log would be:

    [ RUN      ] rv64ua_p_lrsc
      a0 = 0x80002008
      tohost = 0x53b
      An exception occurred.
    [  FAILED  ] rv64ua_p_lrsc
    [==========] 71 test from riscv-tests ran
    [  PASSED  ] 55 test
    [  FAILED  ] 16 test, listed below:
    [  FAILED  ] rv64ui_p_fence_i
    [  FAILED  ] rv64ua_p_amoand_d

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,42 @@ $ make riscv-tests
 ```
 
 You can check the generated report as following:
+```shell
+[==========] Running 71 test(s) from riscv-tests.
+[ RUN      ] rv64ui_p_add
+[       OK ] rv64ui_p_add
+[ RUN      ] rv64ui_p_addi
+[       OK ] rv64ui_p_addi
+
+...
+
+[ RUN      ] rv64ua_p_lrsc
+  a0 = 0x80002008
+  tohost = 0x53b
+  An exception occurred.
+[  FAILED  ] rv64ua_p_lrsc
+[==========] 71 test(s) from riscv-tests ran.
+[  PASSED  ] 55 test(s).
+[  FAILED  ] 16 test(s), listed below:
+[  FAILED  ] rv64ui_p_fence_i
+[  FAILED  ] rv64ua_p_amoand_d
+...
 ```
-=======================
-Test result: 55/71
-=======================
+
+If you want to execute a specific test instead of the whole test, please run the following command:
+```shell
+$ ./semu --test <test_case_name>
+```
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -52,10 +52,42 @@ $ make riscv-tests
...
+[  FAILED  ] rv64ua_p_lrsc
+[==========] 71 test(s) from riscv-tests ran.
+[  PASSED  ] 55 test(s).
+[  FAILED  ] 16 test(s), listed below:
+[  FAILED  ] rv64ui_p_fence_i
+[  FAILED  ] rv64ua_p_amoand_d
+...
+```
+
+If you want to run the specific test, please run the following command:
```
jserv, line 77: Change the statement to
>  execute a specific test instead of the whole test

tests/test.h
```diff
@@ -5,20 +5,24 @@
 #define TEST_FAIL -1
 
 struct testdata {
+    int tested;
```
jserv, line 8: Can you combine `tested` and `result` by bitwise operations? It seems to be straightforward to filter with bitmask(s).

gagachang: Sorry I was confused. We don't need two variables to record the test.
Zero indicates that the test is not ran. Otherwise the test is either passed or failed.
I think this is enough.

tests/test.h
```diff
@@ -1,8 +1,9 @@
 #ifndef __TEST_H__
 #define __TEST_H__
 
-#define TEST_PASS 0
 #define TEST_FAIL -1
+#define NOT_TESTED 0
```
jserv, line 5: Replace `NOT_TESTED` with `UNTESTED`. And use `enum`.

jserv: Alternatively, enforce consistent naming conventions. i.e.,
```c
enum {
    TEST_Unknown = 0,
    TEST_Pass = 1,
    TEST_Fail = -1,
};
```

gagachang: OK Done!

jserv:

PR #12, opened 2022-11-06
Implement more RV64M instructions

gagachang:
Implements more RV64M instructions and adds corresponding tests. Now the result of tests becomes:

    [ RUN      ] rv64ua-p-lrsc
      a0 = 0x80002008
      tohost = 0x53b
      An exception occurred.
    [  FAILED  ] rv64ua-p-lrsc
    [==========] 83 test(s) from riscv-tests ran.
    [  PASSED  ] 67 test(s).
    [  FAILED  ] 16 test(s), listed below:
    [  FAILED  ] rv64ui-p-fence_i
    [  FAILED  ] rv64ua-p-amoand_d

### DIFF
--- a/semu.c
+++ b/semu.c
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include "mul128.h"
 
 /* xv6 uses only 128MiB of memory. */
 #define RAM_SIZE (1024 * 1024 * 128)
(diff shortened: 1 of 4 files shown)


semu.c
```diff
@@ -1023,21 +1023,67 @@ exception_t cpu_execute(struct cpu *cpu, const uint64_t insn)
             cpu->regs[rd] = cpu->regs[rs1] - cpu->regs[rs2];
         } else if (funct3 == 0x1 && funct7 == 0x00) { /* sll */
             cpu->regs[rd] = cpu->regs[rs1] << shamt;
+        } else if (funct3 == 0x1 && funct7 == 0x01) { /* mulh */
+#if __SIZEOF_INT128__
```
jserv, line 1027: `__SIZEOF_INT128__` is defined by Clang and GCC when `__int128` is
supported. For the sake of portability, we shall rewrite as following:
```c
#if defined(__SIZEOF_INT128__) && __SIZEOF_INT128__
...
```

jserv: By the way, there should be a deterministic path when `__int128` is absent.

gagachang: > `__SIZEOF_INT128__` is defined by Clang and GCC when `__int128` is supported. For the sake of portability, we shall rewrite as following:
> 
> ```c
> #if defined(__SIZEOF_INT128__) && __SIZEOF_INT128__
> ...
> ```

done!

gagachang: > By the way, there should be a deterministic path when `__int128` is absent.

I found some [materials from stackoverflow](https://stackoverflow.com/questions/22845801/32-bit-signed-integer-multiplication-without-using-64-bit-data-type) and applied them into **mul128.h**, is it OK for us?

PR #15, opened 2022-11-07
Check address alignment of AMO instructions

gagachang:
For AMOs, the A extension requires that the address held in rs1 be naturally aligned to the size of the operand (i.e., eight-byte aligned for 64-bit words and four-byte aligned for 32-bit words). If the address is not naturally aligned, an address-misaligned exception or an access-fault exception will be generated.

### DIFF
--- a/semu.c
+++ b/semu.c
@@ -115,6 +115,12 @@ typedef enum {
         _a < _b ? _a : _b;      \
     })
 
+/* Check alignement of the address (Assume alignment is a power of 2.)
+ * x is the address to be checked.
+ * a is the alignment constaint and must be power of 2.
+ */
+#define IS_ALIGNED(x, a) (((x) & ((typeof(x)) (a) -1)) == 0)
+
 bool exception_is_fatal(const exception_t e)
 {
     switch (e) {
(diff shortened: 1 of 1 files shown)


semu.c
```diff
@@ -115,6 +115,11 @@ typedef enum {
         _a < _b ? _a : _b;      \
     })
 
+static inline bool is_aligned(const uint64_t addr, size_t byte_count)
+{
+    return ((addr % byte_count) == 0);
```
jserv, line 120: Remove the outer parentheses since they are not necessary.

jserv: Can you use macro `IS_ALIGNED` in Linux kernel? That is,
```c
#define IS_ALIGNED(x, a)        (((x) & ((typeof(x))(a) - 1)) == 0)
```

Here `a` is a power of 2.

gagachang: > Can you use macro `IS_ALIGNED` in Linux kernel? That is,
> 
> ```c
> #define IS_ALIGNED(x, a)        (((x) & ((typeof(x))(a) - 1)) == 0)
> ```
> 
> Here `a` is a power of 2.

Done, I have submited new commit

PR #17, opened 2023-06-17
Handle UART input with termios

JiggerChuang:
The original keybord input handling in UART generates an extra endline character, which causes tab completion and vi editor fauilure.

This change incorparates termios function to address the above issue. Following this change, users can utilize the vi editor, benefit from tab completion, and exit semu by pressing Ctrl-a followed by x.

### DIFF
--- a/uart.c
+++ b/uart.c
@@ -1,7 +1,9 @@
 #include <errno.h>
 #include <poll.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
+#include <termios.h>
 #include <unistd.h>
 
 #include "device.h"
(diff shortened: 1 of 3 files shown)


jserv (changes requested):
Read [CONTRIBUTING.md](https://github.com/jserv/semu/blob/master/CONTRIBUTING.md) carefully and rework the proposed changes.
> Use [snakecase](https://en.wikipedia.org/wiki/Snake_case). Do not use "camelcase".

uart.c
```diff
@@ -12,6 +14,27 @@
...
+static void ResetKeyboardInput()
+{
+    /* Re-enable echo, etc. on keyboard. */
+    struct termios term;
+    tcgetattr(0, &term);
+    term.c_lflag |= ICANON | ECHO;
+    tcsetattr(0, TCSANOW, &term);
+}
+
+/* Override keyboard, so we can capture all keyboard input for the VM. */
```
jserv, line 26: Improve the comments.

jserv:
Known issue: During the booting process of the Linux kernel, the key sequence Ctrl-A X may not have an immediate effect because the emulator itself fails to accept UART events.
@JiggerChuang, add the known issue at the end of git commit message and improve it in another pull request.

jserv:
Thank @JiggerChuang for contributing!

PR #18, opened 2023-07-13
Replace fread with mmap

zoanana990:
Replace `fread` with `mmap` method

Performance comparison
Load Linux image:
- fread method:  18446744073709535686 clock
- mmap method:                                       157 clock

### DIFF
--- a/main.c
+++ b/main.c
@@ -3,6 +3,10 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <time.h>
 
 #include "device.h"
 #include "riscv.h"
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -3,6 +3,10 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <time.h>
```
jserv, line 9: The inclusion `time.h` is redundant.

jserv (changes requested):
Use [atexit(3)](https://man7.org/linux/man-pages/man3/atexit.3.html) to perform the `munmap` operation.

jserv:
Check https://github.com/jserv/semu/blob/master/CONTRIBUTING.md carefully and ensure the consistent style.

jserv:
Any update?

zoanana990:
yes, coding style was changed, but  i don't know why i need to use atexit(3) to perform the munmap operation, is this for error handling? like the following code:

```c
void cleanup(*addr, size_t len)
{
    munmap(...)
}
static void read_file_into_ram(char **ram_loc, const char *name)
{
...
    if (*ram_loc == MAP_FAILED) {
        fprintf(stderr, "mmap failed\n");
        atexit(cleanup);
    }
}
```

jserv:
> yes, coding style was changed, but i don't know why i need to use atexit(3) to perform the munmap operation, is this for error handling? like the following code:

No, `atexit` is invoked right after `mmap` succeeds. It is used for normal execution path rather than the abnormal one. i.e., exception handling.

jserv:
Superseded by commit 944a51d612aef623b51ebcdd4956336bae61e0a6

jserv:
Thank @zoanana990 for contributing.

PR #20, opened 2023-07-22
Handle misalignment properly in jump-link instruction

chiangkd:
From spec, an instruction-address-misaligned exception is reported on the branch or jump instruction, implying rd would not be updated.

For simplicity, I keep `do_jump` function to handle branch instruction, and modify only `op_jump_link` function.

Reference: https://riscv.org/technical/specifications/ (section 2.2)

### DIFF
--- a/riscv.c
+++ b/riscv.c
@@ -677,9 +677,12 @@ static void do_jump(vm_t *vm, uint32_t addr)
 
 static void op_jump_link(vm_t *vm, uint32_t insn, uint32_t addr)
 {
-    /* TODO: If the jump fails (misalign), shall rd have to be updated? */
-    set_dest(vm, insn, vm->pc);
-    do_jump(vm, addr);
+    if (unlikely(addr & 0b11)) {
+        vm_set_exception(vm, RV_EXC_PC_MISALIGN, addr);
+    } else {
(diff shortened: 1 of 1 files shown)


riscv.c
```diff
@@ -677,9 +677,12 @@ static void do_jump(vm_t *vm, uint32_t addr)
 
 static void op_jump_link(vm_t *vm, uint32_t insn, uint32_t addr)
 {
-    /* TODO: If the jump fails (misalign), shall rd have to be updated? */
-    set_dest(vm, insn, vm->pc);
-    do_jump(vm, addr);
+    if (unlikely(addr & 0b11))
```
jserv, line 680: Rewrite as following:
```c
    if (unlikely(addr & 0b11)) {
        vm_set_exception(vm, RV_EXC_PC_MISALIGN, addr);
    } else {
        set_dest(vm, insn, vm->pc);
        vm->pc = addr;
    }
```

jserv (changes requested):
Use ChatGPT to refine / proofread git commit messages.

jserv:
In addition, use plain text in git commit messages rather than markdown.

jserv:

Issue #24, opened 2023-07-27
Implement the loading of initrd image


jserv:
At present, initrd is missing, which means users have to boot Linux with initramfs enabled. This might not be convenient for someone who expects to specify a customized userland and/or benchmark suite at an early stage.

Expected output:
1. specify `initrd-start` and `initrd-end` in device tree.
2. Add an option `-i` to specify initrd image along with the functionality to load.

Reference:
* [yariscv](https://github.com/fgssfgss/yariscv)
* [kvm-host](https://github.com/sysprog21/kvm-host)

jserv:
Close via commit [61cea98](https://github.com/sysprog21/semu/commit/61cea98f5b5e43216c9ae00aa481a7dc940296da).

PR #25, opened 2023-07-27
Implement the loading of initrd image

shengwen-tw:
This pull request is for solving #24, where the modifications mainly include:

1. Reserve 8 MiB of memory at the end of the guest OS before the dtb region, and map initrd image to the new reserved area.
2. Adjust device tree source by specifying `initrd-start` and `initrd-end` properties.
3. Support passing initrd image with -i option.
4. Configure the Linux kernel to not let it be integrated with `initramfs` by default.
5. Adjust `build-image.sh` to copy the `rootfs.cpio` file from `buildroot/` to project's root directory.


main.c
```diff
@@ -371,13 +378,27 @@ static int semu_start(int argc, char **argv)
     }
     assert(!(((uintptr_t) emu.ram) & 0b11));
 
+    /* *-----------------------------------------*
+     * |              Memory layout              |
+     * *--------------*----------------*---------*
+     * | kernel image |  initrd image  |   dtb   |
+     * *--------------*----------------*---------*
+     * <-   501MB   ->*<-    10MB    ->*<- 1MB ->*
```
jserv, line 386: I am not sure if it still makes sense to preserve large space for kernel image since initrd can be specified on demand.

shengwen-tw: Currently the `rootfs.cpio` file by the Buildroot is about 3.4MiB.
How large of the memory do you suggest to reserve for the initrd image?

jserv: cpio might be uncompressed, and its size is typically larger on desktop. e.g., more than 100 MB for Ubuntu Linux. Here, I suggest to preserve 32MB (see [New Binary Format](https://man.archlinux.org/man/cpio.5.en#New_Binary_Format)).

shengwen-tw:
Reservation of 32MiB introduces about 20~30 seconds of extra booting time on my `AMD Ryzen 7 PRO 4750U (16 cores)` machine.

The user may find the OS stop at the following line for a while:
```
[    0.152986] Unpacking initramfs...
```

One dirty fix for short term might be shrinking the size of the `linux,initrd-end` property for now.

jserv:
> Reservation of 32MiB introduces about 20~30 seconds of extra booting time on my `AMD Ryzen 7 PRO 4750U (8 cores)` machine.

Let's set 8 MB by default and introduce build-time option to specify the memory layout.

main.c
```diff
@@ -371,13 +378,26 @@ static int semu_start(int argc, char **argv)
...
     char *ram_loc = (char *) emu.ram;
     /* Load Linux kernel image */
     map_file(&ram_loc, kernel_file);
-    /* Load at last 1 MiB to prevent kernel / initrd from overwriting it */
-    uint32_t dtb_addr = RAM_SIZE - 1024 * 1024; /* Device tree */
+    /* Load at last 1 MiB to prevent kernel from overwriting it */
+    uint32_t dtb_addr = RAM_SIZE - DTB_SIZE; /* Device tree */
     ram_loc = ((char *) emu.ram) + dtb_addr;
     map_file(&ram_loc, dtb_file);
+    /* Load optional initrd image at last 10 Mib before the dtb region to
```
jserv, line 394: 8 MiB?

PR #27, opened 2023-08-22
Unset CONFIG_STRICT_KERNEL_RWX in kernel configuration

onnokort:
This will reduce kernel size from ~21MiB to ~5MiB as it removes huge alignment sections in the linker script.
For details, see symbol SECTION_ALIGN in

arch/riscv/include/asm/set_memory.h

and

arch/riscv/kernel/vmlinux.lds.S

It should be noted, however, that this might trade kernel binary load size for security in some settings.


jserv:
Thank @onnokort for contributing!

Issue #26, opened 2023-08-21
Route MMU through RAM access methods, caching?


onnokort:
Hi,

I just ported your great little emu to a [very unlikely architecture](https://github.com/onnokort/semu-c64/) and whilst doing so, I had to tweak the MMU code to go through the RAM access methods as well. I propose the code would be a lot more portable if MMU RAM access calls corresponding methods/functions as well, so that RAM access can really be anything. The current way of having a pointer into emu RAM is a bit of a breach of abstraction IMO.

I further noticed that every CPU instruction fetch goes through the MMU page walk and a very simple "1-entry-1-page" cache (one for INSN fetch, load & store) speeds up the emulation considerably. For the standard PC build, I got from about 10s to boot the emulated Linux to about 7s just by caching the current page for `mmu_fetch`.

As my code to do so is quite peculiar and adapted to the odd target architecture, I don't think it really fits well as an upstream patch. Still, I like to raise the issue of the MMU being direct-access here.

jserv:
@RinHizakura, Could you provide a comment on the MMU issue mentioned above, drawing from your past experience in developing the [riscv-emulator](https://github.com/RinHizakura/riscv-emulator)?

RinHizakura:
For the part of the instruction cache, it can really help a lot for system emulators. In my experience, I have  [icache.c](https://github.com/RinHizakura/riscv-emulator/blob/main/src/icache.c) on my risc-v emulator, and it improves performance even though the implementation is super naive as you can see. [riscv-rust](https://github.com/takahirox/riscv-rust) has a better implementation of page cache if we want one to reference.

In conclusion, the instruction cache would be an important component for optimization, but it is also worth mentioning that the timing to invalidate the cache for system emulator should be considered carefully for correctness.

jserv:
> I further noticed that every CPU instruction fetch goes through the MMU page walk and a very simple "1-entry-1-page" cache (one for INSN fetch, load & store) speeds up the emulation considerably. For the standard PC build, I got from about 10s to boot the emulated Linux to about 7s just by caching the current page for `mmu_fetch`.
> 
> As my code to do so is quite peculiar and adapted to the odd target architecture, I don't think it really fits well as an upstream patch. Still, I like to raise the issue of the MMU being direct-access here.

I appreciate the outstanding C64 porting work carried out by @onnokort! Introducing a cache for MMU manipulation is a logical step towards enhancing performance.

[...]

PR #28, opened 2023-08-24
Caching for MMU translations

onnokort:
Here is the *proposed* MMU caching scheme as discussed also here: https://github.com/sysprog21/semu/issues/26

This is likely not yet final and needs to be discussed. As said, the savings at -O3 are not great, YMMV. If you add extra test delays to RAM access, you can hopefully see that this does as it should.

Given that the MMU logic is one of, if not the most involved parts of the code, I have a couple of questions & doubts about my own code still:

- there is a separate patch for adding a #cycles-run limit to the emulator for performance testing
- I have written an adhoc emulator RAM dump at the end (not in this PR) and did a bit for bit comparison and it looks like the cache does not change anything it terms of how the current Linux image boots (first 300Mcycles of the current kernel + basic buildroot till login prompt on a 16MiB RAM `semu`)
- I have noticed that the RISCV MMU spec allows for an even simpler MMU implementation in the single-core case, which does not update any access bits upon lookup. More specifically, I have noticed that if I comment out this part from riscv.c:
```
if (new_pte != pte)
        *pte_ref = new_pte;
```
that nothing changes in terms of runtime behaviour (again, did a bit-for-bit check of the output RAM at the end of ~300Mcycles) of the emulator. What gives? Is that to be expected? Is that simply how Linux assumes everything to be if on a single RISCV32 core? If so, is that something that should potentially be conditionally #ifdef-ed out to optimize for single-core use of the emu? I must really admit that I have no great oversight of how it all fits together with the exact Linux MMU accesses yet, I am really a bit out of my depth here at the moment :D

- Likewise, I added a couple of MMU cache invalidations wherever I saw the potential need for them and I assumed that they are necessary (so as to not expose kernel translations to user space, for example) whenever the execution mode changes between user and supervisor mode. Is my set of invalidations sufficient? Is it too much? Or actually correct?

- (As an aside: having the EMU exported to python or another scripting language would make all the high level stuff like cmdline parsing and such test jigs for e.g. a regression test doing RAM bit-for-bit checks against a 'golden RAM dump' *so* much simpler ...)


mmu_cache.h
```diff
@@ -0,0 +1,65 @@
...
+    MMU_CACHE_IDX_TYPE write_idx;
+    uint32_t from[MMU_CACHE_ENTRIES];
+    uint32_t to[MMU_CACHE_ENTRIES];
+#if MMU_CACHE_COUNTERS
+    uint64_t hits, misses;
+#endif
+};
+
+static inline void mmu_cache_reset_ctx(struct _mmu_cache_ctx* cc) {
+    if (MMU_CACHE_ENTRIES) {
```
jserv, line 22: Rewrite as following:
```c
#if MMU_CACHE_ENTRIES > 0
    cc->write_idx = 0;
    ...
#endif
```

mmu_cache.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+#include <string.h>
+/* This currently implements a simple, fixed-size LRU cache for MMU
+   entries. Assumes that address 0 is never cached and it is used to
+   flag empty entries. If a returned translation is zero, the calling
+   logic has to assume that this means 'no entry available'. */
+
+#define MMU_CACHE_ENTRIES 3
+#define MMU_CACHE_IDX_TYPE int
+#define MMU_CACHE_COUNTERS 0
```
jserv, line 10: How about `MMU_CACHE_STAT`, which stands for statistics?

jserv:
> I have noticed that the RISCV MMU spec allows for an even simpler MMU implementation in the single-core case, which does not update any access bits upon lookup. More specifically, I have noticed that if I comment out this part from riscv.c:
> ```c
> if (new_pte != pte)
>         *pte_ref = new_pte;
> ```
> that nothing changes in terms of runtime behaviour (again, did a bit-for-bit check of the output RAM at the end of ~300Mcycles) of the emulator. What gives? Is that to be expected? Is that simply how Linux assumes everything to be if on a single RISCV32 core? If so, is that something that should potentially be conditionally #ifdef-ed out to optimize for single-core use of the emu? I must really admit that I have no great oversight of how it all fits together with the exact Linux MMU accesses yet, I am really a bit out of my depth here at the moment.

Given the current heavily simplified MMU implementation, the page table walker might not even traverse into such cases, so I could potentially use tools like stress-ng to generate memory stressing workloads and determine the necessity. Initially, we should consider leaving some FIXME/TODO comments in the source files. Would you agree?

jserv (changes requested):
Check CONTRIBUTING.md and run `clang-format` to ensure the consistent style.

PR #29, opened 2023-08-30
Fix signed division overflow case

fourcolor:
Refer to [RISC-V ISA spec](https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf) Chapter 6,  signed division overflow occurs only when the most-negative integer, $−2^{XLEN −1}$, is divided by $−1$. The quotient of signed division overflow is equal to the dividend, and the remainder is zero.


jserv:
Thank @fourcolor for contributing!

PR #35, opened 2024-01-08
Fix signed integer overflow in RV32M

visitorckw:
The current implementation of the mul instruction does not guard against integer overflow, potentially leading to undefined behavior. Cast the operands to int64_t before performing the multiplication to ensure that the result can be accommodated without overflow. The lower 32 bits of the product are then extracted, preserving the correct uint32_t type.


jserv:

PR #37, opened 2024-01-14
Suppress DTC warnings

ChinYikMing:
The DTC of version 1.6.1 generates the following warnings:
1. (unit_address_vs_reg) /soc: node has a reg or ranges property, but no unit name
2. (interrupt_provider): /cpus/cpu@0/interrupt-controller: Missing #address-cells in interrupt provider

To suppress them, add unit-name with parent-bus-address to the SOC node and #address-cells to the CPU interrupt controller.

### DIFF
--- a/minimal.dts
+++ b/minimal.dts
@@ -6,7 +6,7 @@
     model = "semu";
 
     aliases {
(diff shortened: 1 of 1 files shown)


jserv:

PR #43, opened 2024-05-15
Bump buildroot and binutils version to stable tag

ChinYikMing:
The stable tag of buildroot is 2024.02.x currently. Thus, upgrade the default buildroot version of 2023.11.x to 2024.02.x .

The commit 1391c99 in branch 2024.02.x of buildroot dropped the support of 2.38.x series of binutils and set the default version to 2.40.x. To ensure the compliance, upgrade the default configuration of buildroot as well.

This commit also fix the bumping to the wrong version of binutils in commit 2c1e29c.


ChinYikMing:
@shengwen-tw , can you help to verify?

jserv:
I defer to @shengwen-tw for confirmation.

shengwen-tw:
I confirmed that Buildroot with updated `2024.02.x` tag could indeed enable `binutils` with default version of 2.40. (Though 2.41 is also feasible.)

jserv:
> I confirmed that Buildroot with updated `2024.02.x` tag could indeed enable `binutils` with default version of 2.40. (Though 2.41 is also feasible.)

binutils-2.41 should be selected rather than the proposed `2.40`.

ChinYikMing:
> > I confirmed that Buildroot with updated `2024.02.x` tag could indeed enable `binutils` with default version of 2.40. (Though 2.41 is also feasible.)
> 
> binutils-2.41 should be selected rather than the proposed `2.40`.

Updated.

PR #44, opened 2024-05-16
Decouple interrupt bit index calculation from ilog2

ChinYikMing:
ilog2 is a common function, but interrupt bit index calculation is application dependent. Thus, decouple them and create a new macro called GET_INTR_IDX which leverages the ilog2 to calculate the interrupt bit index.

### DIFF
--- a/common.h
+++ b/common.h
@@ -14,8 +14,19 @@
 /* Ensure that __builtin_clz is never called with 0 argument */
 static inline int ilog2(int x)
 {
-    return 31 - __builtin_clz(x | 1);
+    return __builtin_clz(x | 1);
 }
+
+/*
+ * Note: only applicable to 32-bit number
+ *
+ * Example output:
+ * GET_INTR_IDX(1) = 0
+ * GET_INTR_IDX(2) = 1
+ * GET_INTR_IDX(4) = 2
+ * GET_INTR_IDX(8) = 3
+ */
+#define GET_INTR_IDX(x) (31 - ilog2(x))
 
 /* Range check
  * For any variable range checking:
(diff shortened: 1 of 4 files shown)


common.h
```diff
@@ -14,9 +14,20 @@
 /* Ensure that __builtin_clz is never called with 0 argument */
 static inline int ilog2(int x)
 {
-    return 31 - __builtin_clz(x | 1);
+    return __builtin_clz(x | 1);
```
jserv, line 17: floor(log2(x)) should be `31 - __builtin_clz(x | 1)`.

ChinYikMing: > floor(log2(x)) should be `31 - __builtin_clz(x | 1)`.

`31 - __builtin_clz(x | 1)` seems like a calculation of bit set in a 32-bit number instead of floor(log2(x)) ?

jserv: Check this: https://github.com/apple/swift-llbuild/blob/main/include/llvm/Support/MathExtras.h#L536-L541

common.h
```diff
@@ -14,9 +14,20 @@
...
+/*
+ * Note: only applicable to 32-bit number
+ *
+ * Example output:
+ * GET_INTR_IDX(1) = 0
+ * GET_INTR_IDX(2) = 1
+ * GET_INTR_IDX(4) = 2
+ * GET_INTR_IDX(8) = 3
+ */
+#define GET_INTR_IDX(x) (31 - ilog2(x))
```
jserv, line 29: The naming is confusing. Refine it.

common.h
ChinYikMing: > Check this: https://github.com/apple/swift-llbuild/blob/main/include/llvm/Support/MathExtras.h#L536-L541

Seems like I had misunderstood it. I would close this PR.

PR #46, opened 2024-06-15
Preliminary SMP support

ranvd:
In this commit, semu is able to simulate SMP architecture running on the Linux kernel.

Before simulating SMP on semu, we need to enable SMP support in the Linux kernel. Please cross-compile the Linux kernel with the configuration file located at configs/linux.config.

After recompiling the Linux kernel with SMP support enabled, simply execute `make check SMP=1` to simulate a quad-hart RISC-V CPU.

```
processor	: 0
hart		: 0
isa		: rv32ima_zicntr_zicsr_zifencei_zihpm
mmu		: sv32
mvendorid	: 0x12345678
marchid		: 0x80000001
mimpid		: 0x1
hart isa	: rv32ima_zicntr_zicsr_zifencei_zihpm

processor	: 1
hart		: 1
isa		: rv32ima_zicntr_zicsr_zifencei_zihpm
mmu		: sv32
mvendorid	: 0x12345678
marchid		: 0x80000001
mimpid		: 0x1
hart isa	: rv32ima_zicntr_zicsr_zifencei_zihpm

processor	: 2
hart		: 2
isa		: rv32ima_zicntr_zicsr_zifencei_zihpm
mmu		: sv32
mvendorid	: 0x12345678
marchid		: 0x80000001
mimpid		: 0x1
hart isa	: rv32ima_zicntr_zicsr_zifencei_zihpm

processor	: 3
hart		: 3
isa		: rv32ima_zicntr_zicsr_zifencei_zihpm
mmu		: sv32
mvendorid	: 0x12345678
marchid		: 0x80000001
mimpid		: 0x1
hart isa	: rv32ima_zicntr_zicsr_zifencei_zihpm
```

PR #47, opened 2024-06-24
Fix incompatible mmu-type

chiangkd:
Per https://github.com/riscv-non-isa/riscv-device-tree-doc/blob/master/bindings/riscv/cpus.txt , the `mmu-type` attribute should be set to "riscv,sv32"


ranvd:
This change work fine on #46 .

jserv:

PR #48, opened 2024-06-27
Implement LRU TLB

nelson0720j:
Implement LRU TLB with doubly link list. 
However, the error of not being able to execute correctly has not been resolved yet.

### DIFF
--- a/riscv.c
+++ b/riscv.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <stdlib.h>
 
 #include "riscv.h"
 #include "riscv_private.h"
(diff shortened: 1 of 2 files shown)


riscv.c
```diff
@@ -170,6 +171,21 @@ static inline uint32_t read_rs2(const vm_t *vm, uint32_t insn)
 static void mmu_invalidate(vm_t *vm)
 {
     vm->cache_fetch.n_pages = 0xFFFFFFFF;
+
+    struct node_t *current = vm->tlb_list.head;
```
jserv, line 175: Replace while loop with for iteration, leading the shorter code. In addition, the condition `current != NULL` can be simplified as `current`.

riscv.c
```diff
@@ -234,6 +250,76 @@ static bool mmu_lookup(const vm_t *vm,
...
+    node->next = tmp_next;
+    tmp_next->prev = node;
+}
+
+/* search TLB */
+static uint32_t *tlb_lookup(vm_t *vm, uint32_t vpn, bool write)
+{
+    struct node_t *current = vm->tlb_list.head->next;
+
+    while (current != vm->tlb_list.tail) {
```
jserv, line 285: Ditto. For iteration substitution.

riscv.h
```diff
@@ -34,6 +34,26 @@ typedef struct {
...
 
+#define TLB_SIZE 16
+typedef struct {
+    uint32_t vpn;
+    uint32_t *pte;
+    bool valid;
+    bool dirty;
+} tlb_entry_t;
+
+struct node_t{
```
jserv, line 45: Always run `clang-format -i` before submitting changes.

jserv:
Have you figured out the root cause of failing to execute init process?

nelson0720j:
not yet, I am still working on it.


Jim Huang ***@***.***> 於 2024年7月6日 週六 下午11:21寫道：

PR #49, opened 2024-07-14
Replace mtime counter with realtime timer

chiangkd:
Replace `mtime` counter by retrieving time of the specified clock ID. Also, Real time counter (RTC) mechanism is implemented to prevent `mtime` overflow.

Before:
```
[    0.000000] Linux version 6.1.94 (aaron@aaron-Lenovo-Y520-15IKBN) (riscv32-buildroot-linux-gnu-gcc.br_real (Buildroot 2024.02.3) 12.3.0, GNU ld (GNU Binutils) 2.41) #8 SMP Thu Jul 11 00:37:05 CST 2024
[    0.000000] Machine model: semu
[    0.000000] earlycon: ns16550 at MMIO 0xf4000000 (options '')
[    0.000000] printk: bootconsole [ns16550] enabled
[    0.000000] Zone ranges:
[    0.000000]   Normal   [mem 0x0000000000000000-0x000000001fffffff]
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
...
[    2.687661] Freeing initrd memory: 8188K
[    2.691697] Freeing unused kernel image (initmem) memory: 180K
[    2.691978] Kernel memory protection not selected by kernel config.
[    2.692277] Run /init as init process

```
Now:
```
[    0.000000] Linux version 6.1.94 (aaron@aaron-Lenovo-Y520-15IKBN) (riscv32-buildroot-linux-gnu-gcc.br_real (Buildroot 2024.02.3) 12.3.0, GNU ld (GNU Binutils) 2.41) #10 SMP Mon Jul 15 02:47:57 CST 2024
[    0.000000] Machine model: semu
[    0.000000] earlycon: ns16550 at MMIO 0xf4000000 (options '')
[    0.000000] printk: bootconsole [ns16550] enabled
[    0.000000] Zone ranges:
[    0.000000]   Normal   [mem 0x0000000000000000-0x000000001fffffff]
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
...
[   73.908933] Freeing initrd memory: 8188K
[   74.033405] Freeing unused kernel image (initmem) memory: 180K
[   74.044397] Kernel memory protection not selected by kernel config.
[   74.064088] Run /init as init process
```
Revised boot log to display actual timer during boot flow.

Reference to RVVM:https://github.com/LekKit/RVVM/blob/staging/src/rvtimer.c

However, another issue accopanies this PR
```
[   32.038173] rcu: INFO: rcu_sched self-detected stall on CPU
[   32.043867] rcu: 	3-....: (5248 ticks this GP) idle=846c/1/0x40000002 softirq=371/371 fqs=2114
[   32.055848] 	(t=5251 jiffies g=-883 q=10 ncpus=4)
[   32.064761] CPU: 3 PID: 39 Comm: kworker/u8:1 Not tainted 6.1.94 #10
```
RCU issues a CPU stall wrning when it waits more than a grace period, which is normally 21 seconds.

Detail here: https://docs.kernel.org/RCU/stallwarn.html

### DIFF
--- a/clint.c
+++ b/clint.c
@@ -5,7 +5,9 @@
 
 void clint_update_interrupts(hart_t *hart, clint_state_t *clint)
 {
-    if (clint->mtime > clint->mtimecmp[hart->mhartid])
+    uint64_t time_delta =
+        clint->mtimecmp[hart->mhartid] - semu_timer_get(&hart->time);
+    if ((int64_t) time_delta <= 0)
         hart->sip |= RV_INT_STI_BIT;
     else
         hart->sip &= ~RV_INT_STI_BIT;
@@ -31,7 +33,8 @@ static bool clint_reg_read(clint_state_t *clint, uint32_t addr, uint32_t *value)
     }
 
     if (addr < 0xBFFF) {
-        *value = clint->mtime >> (32 & -!!(addr & 0b100));
+        *value = (uint32_t) (semu_timer_get(&clint->mtime) >>
+                             (32 & -!!(addr & 0b100)));
         return true;
     }
     return false;
@@ -58,14 +61,14 @@ static bool clint_reg_write(clint_state_t *clint, uint32_t addr, uint32_t value)
     }
 
     if (addr < 0xBFFF) {
-        int32_t upper = clint->mtime >> 32;
-        int32_t lowwer = clint->mtime;
+        int32_t upper = clint->mtime.begin >> 32;
+        int32_t lower = clint->mtime.begin;
         if (addr & 0b100)
             upper = value;
         else
-            lowwer = value;
+            lower = value;
 
-        clint->mtime = (uint64_t) upper << 32 | lowwer;
+        semu_timer_rebase(&clint->mtime, (uint64_t) upper << 32 | lower);
         return true;
     }
     return false;
--- a/main.c
+++ b/main.c
@@ -81,13 +81,6 @@ static void emu_update_timer_interrupt(hart_t *hart)
     clint_update_interrupts(hart, &data->clint);
 }
 
-static void emu_update_global_timer(vm_t *vm)
-{
-    emu_state_t *data = PRIV(vm->hart[0]);
-    data->clint.mtime++;
-    return;
-}
-
 static void mem_load(hart_t *hart,
                      uint32_t addr,
                      uint8_t width,
@@ -523,6 +516,7 @@ static int semu_start(int argc, char **argv)
     /* Initialize the emulator */
     emu_state_t emu;
     memset(&emu, 0, sizeof(emu));
+    semu_timer_init(&emu.clint.mtime, CLOCK_FREQ);
 
     /* Set up RAM */
     emu.ram = mmap(NULL, RAM_SIZE, PROT_READ | PROT_WRITE,
@@ -591,8 +585,6 @@ static int semu_start(int argc, char **argv)
     /* Emulate */
     uint32_t peripheral_update_ctr = 0;
     while (!emu.stopped) {
-        emu_update_global_timer(&vm);
-
         for (uint32_t i = 0; i < vm.n_hart; i++) {
             if (peripheral_update_ctr-- == 0) {
                 peripheral_update_ctr = 64;
--- a/utils.h
+++ b/utils.h
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <stdint.h>
+
+/* TIMER */
+typedef struct {
+    uint64_t begin;
+    uint64_t freq;
+} semu_timer_t;
+
+void semu_timer_init(semu_timer_t *timer, uint64_t freq);
+uint64_t semu_timer_get(semu_timer_t *timer);
+void semu_timer_rebase(semu_timer_t *timer, uint64_t time);
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,11 @@ CC ?= gcc
 CFLAGS := -O2 -g -Wall -Wextra
 CFLAGS += -include common.h
 
+# clock frequency
+CLOCK_FREQ ?= 65000000
+DT_CFLAGS := -D CLOCK_FREQ=$(CLOCK_FREQ)
+CFLAGS += $(DT_CFLAGS)
+
 OBJS_EXTRA :=
 # command line option
 OPTS :=
@@ -42,6 +47,7 @@ all: $(BIN) minimal.dtb
 OBJS := \
 	riscv.o \
 	ram.o \
+	utils.o \
 	plic.o \
 	uart.o \
 	main.o \
@@ -72,11 +78,12 @@ S := $E $E
 SMP ?= 1
 .PHONY: riscv-harts.dtsi
 riscv-harts.dtsi:
-	$(Q)python3 scripts/gen-hart-dts.py $@ $(SMP)
+	$(Q)python3 scripts/gen-hart-dts.py $@ $(SMP) $(CLOCK_FREQ)
 
 minimal.dtb: minimal.dts riscv-harts.dtsi
 	$(VECHO) " DTC\t$@\n"
 	$(Q)$(CC) -nostdinc -E -P -x assembler-with-cpp -undef \
+	    $(DT_CFLAGS) \
 	    $(subst ^,$S,$(filter -D^SEMU_FEATURE_%, $(subst -D$(S)SEMU_FEATURE,-D^SEMU_FEATURE,$(CFLAGS)))) $< \
 	    | $(DTC) - > $@
 
--- a/device.h
+++ b/device.h
@@ -175,7 +175,7 @@ uint32_t *virtio_blk_init(virtio_blk_state_t *vblk, char *disk_file);
 typedef struct {
     uint32_t msip[4096];
     uint64_t mtimecmp[4095];
-    uint64_t mtime;
+    semu_timer_t mtime;
 } clint_state_t;
 
 void clint_update_interrupts(hart_t *vm, clint_state_t *clint);
--- a/minimal.dts
+++ b/minimal.dts
@@ -24,7 +24,7 @@
     cpus {
         #address-cells = <1>;
         #size-cells = <0>;
-        timebase-frequency = <65000000>;
+        timebase-frequency = <CLOCK_FREQ>;
     };
 
     sram: memory@0 {
--- a/riscv.c
+++ b/riscv.c
@@ -434,10 +434,10 @@ static void csr_read(hart_t *vm, uint16_t addr, uint32_t *value)
 {
     switch (addr) {
     case RV_CSR_TIME:
-        *value = vm->time;
+        *value = semu_timer_get(&vm->time);
         return;
     case RV_CSR_TIMEH:
-        *value = vm->time >> 32;
+        *value = semu_timer_get(&vm->time) >> 32;
         return;
     case RV_CSR_INSTRET:
         *value = vm->instret;
--- a/riscv.h
+++ b/riscv.h
@@ -3,6 +3,8 @@
 #include <stdbool.h>
 #include <stdint.h>
 
+#include "utils.h"
+
 /* ERR_EXCEPTION indicates that the instruction has raised one of the
  * exceptions defined in the specification. If this flag is set, the
  * additional fields "exc_cause" and "exc_val" must also be set to values
@@ -70,8 +72,7 @@ struct __hart_internal {
      * resets.
      */
     uint64_t instret;
-    uint64_t time;
-
+    semu_timer_t time;
     /* Instruction execution state must be set to "NONE" for instruction
      * execution to continue. If the state is not "NONE," the vm_step()
      * function will exit.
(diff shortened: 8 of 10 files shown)


timer.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+
+/* TIMER */
```
jserv, line 7: Let's put all the definitions about devices such as timer into the header `device.h`.

clint.c
```diff
@@ -5,7 +5,9 @@
 
 void clint_update_interrupts(hart_t *hart, clint_state_t *clint)
 {
-    if (clint->mtime > clint->mtimecmp[hart->mhartid])
+    uint64_t time_delta =
+        clint->mtimecmp[hart->mhartid] - semu_timer_gettime(&hart->time);
+    if (time_delta & 0x8000000000000000 || time_delta == 0)
```
ranvd, line 10: Is it better to make `time_delta` an `int64_t` instead of unsigned integer? By doing this, the comparison in the `if` statement can be reduced and more intuitive.

ranvd:
I believe using a real-time counter is a better approach for simulating the timer device. However, I didn't notice any differences between the "Before" and "Now" sections in your text.

Also, Could you please explain how to reproduce the RCU issues you mentioned? Thank you.

ranvd:
> I believe using a real-time counter is a better approach for simulating the timer device. However, I didn't notice any differences between the "Before" and "Now" sections in your text.

Sorry I need to take back what I said earlier. The real-time counter may not always be a better implementation. After measuring the boot-up time with SMP=4, the original version takes around 23 seconds, while this commit takes 39 seconds. I think this is because the real-time counter generates more interrupts then the original version.

Does the real-time counter (RTC) provide any features that can't be neglected?

clint.c
chiangkd: You are correct, casting `time_delta` can be more intutive. Below is the refined implementation
```c
uint64_t time_delta =
    clint->mtimecmp[hart->mhartid] - semu_timer_gettime(&hart->time);
if ((int32_t)time_delta <= 0)
...
```

chiangkd:
> Does the real-time counter (RTC) provide any features that can't be neglected?

Initially, my intention was to enable the actual clock for the RDTIME (0xC01/0xC81) pseudo-instruction, which originally retrieves the monotonic-incremented `mtime`. I did not anticipate that this PR would trigger RCU warnings prior to implementation. Post-implementation, real timestamps are visible in the booting log due to CONFIG_PRINTK_TIME being enabled by default.

>SEMU becomes slower as the number of simulated harts grows.

I acknowledge the observation that SEMU slows down as the number of simulated harts increases. While reducing the frequency in `semu_timer_init` might mitigate this, it would compromise the provision of an actual clock through the TIMER CSR. In my undetstanding, simulating a multi-core system with a single thread inherently reduces efficiency(include much more timer interrupt and much more `clock_gettime` system call in this PR), leading to RCU warnings.
>Since RVVM is a multi-threaded RISC-V simulator, it's okay to directly simulate the timer as you implemented.

I believe implementing multi-threaded simulation for SEMU could provide a better solution to the RCU warning issue, is that make sense?

timer.h
chiangkd: I chose to separate timer.h because timer devices need to be accessible in both `hart_t` and `clint_state_t`. Should I move its contents to `common.h`?

jserv: > I chose to separate `timer.h` because timer devices need to be accessible in both `hart_t` and `clint_state_t`. Should I move its contents to `common.h`?

Move to `utils.[ch]` to avoid potential confusions. When we say "timer," we refer to the peripherals at first glance.

main.c
```diff
@@ -523,6 +516,7 @@ static int semu_start(int argc, char **argv)
     /* Initialize the emulator */
     emu_state_t emu;
     memset(&emu, 0, sizeof(emu));
+    semu_timer_init(&emu.clint.mtime, 65000000);
```
jserv, line 519: Make it configurable. That is, don't hardcode `65000000`.

jserv (changes requested):
Squash the git commits and rebase the latest `master` branch.

jserv:
Properly rebase the latest `master` branch via the following commands:
```shell
$ git remote add upstream https://github.com/sysprog21/semu
$ git fetch upstream 
$ git rebase upstream/master
$ git push --force origin master
```

Next time, create a dedicated branch rather than `master` branch when you are going to submit pull requests.

chiangkd:
> Next time, create a dedicated branch rather than `master` branch when you are going to submit pull requests.

Thank you for the instructions. I understand and will follow these steps and will also make sure to create a dedicated branch for future PR.

jserv (changes requested):
Improve the git commit message. You should explain the rationale.
See https://cbea.ms/git-commit/

utils.h
```diff
@@ -0,0 +1,18 @@
...
+#define TIMER_FREQUENCY 65000000
+
+/* TIMER */
+typedef struct {
+    uint64_t begin;
+    uint64_t freq;
+} semu_timer_t;
+
+void semu_timer_init(semu_timer_t *timer, uint64_t freq);
+uint64_t semu_timer_clocksource(uint64_t freq);
```
jserv, line 16: It is not necessary to expose `semu_timer_clocksource`. Make it a static function.

utils.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+
+#define TIMER_FREQUENCY 65000000
```
jserv, line 7: Don't specify here. Instead, define in `Makefile`. e.g.,
```
DT_CFLAGS := -D TIMER_FREQ=65000000
CFLAGS += $(DT_CFLAGS)
```

The device tree should be changed accordingly. That is,
```diff
--- a/Makefile
+++ b/Makefile
@@ -70,6 +73,7 @@ S := $E $E
 minimal.dtb: minimal.dts
        $(VECHO) " DTC\t$@\n"
        $(Q)$(CC) -nostdinc -E -P -x assembler-with-cpp -undef \
+           $(DT_CFLAGS) \
            $(subst ^,$S,$(filter -D^SEMU_FEATURE_%, $(subst -D$(S)SEMU_FEATURE,-D^SEMU_FEATURE,$(CFLAGS)))) $< \
            | $(DTC) - > $@
 
--- a/minimal.dts
+++ b/minimal.dts
@@ -19,7 +19,7 @@
     cpus {
         #address-cells = <1>;
         #size-cells = <0>;
-        timebase-frequency = <65000000>;
+        timebase-frequency = <CLOCK_FREQ>;
         cpu0: cpu@0 {
             device_type = "cpu";
             compatible = "riscv";
```

clint.c
```diff
@@ -5,7 +5,9 @@
 
 void clint_update_interrupts(hart_t *hart, clint_state_t *clint)
 {
-    if (clint->mtime > clint->mtimecmp[hart->mhartid])
+    uint64_t time_delta =
+        clint->mtimecmp[hart->mhartid] - semu_timer_gettime(&hart->time);
+    if ((int32_t) time_delta <= 0)
```
ranvd, line 10: I think using `int32_t` might trigger potential errors since this is a real-time counter. We can't predict when the `clint_update_interrupts` function will be executed. The time_delta could be a very small negative value, which would normally trigger the timer interrupt. However, if we cast it into a 32-bit integer, the value could incorrectly become positive. For example, if `time_delta` is 0xFFFFFFFF00000001, which is -4294967295, casting it to `int32_t` will result in the value becoming 1.

Therefore, I suggest using `int64_t` instead.

ranvd, line 10: I think using `int32_t` might trigger potential errors since this is a real-time counter. We can't predict when the `clint_update_interrupts` function will be executed. The time_delta could be a very small negative value, which would normally trigger the timer interrupt. However, if we cast it into a 32-bit integer, the value could incorrectly become positive. For example, if `time_delta` is 0xFFFFFFFF00000001, which is -4294967295, casting it to `int32_t` will result in the value becoming 1.

Therefore, I suggest using `int64_t` instead.

chiangkd: Yes, you're absolutely right. When the larger number exceeds the small one by more than 0x100000000, it will incorrectly cast a negative number to a positive one. Thanks for pointing that out!

utils.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <stdbool.h>
```
jserv, line 3: Don't include `<stdbool.h>` since it is not used.

utils.h
```diff
@@ -0,0 +1,15 @@
...
+#include <stdint.h>
+
```
jserv, line 5: Keep only one blank line.

jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully and refine your git commit messages.
> Wrap the body at 72 characters

Avoid using [backticks](https://en.wikipedia.org/wiki/Backtick) in git commit messages and comments to ensure compatibility with terminal emulators that may not render the backtick character properly.

Instead, use the pair `'` or `"` (quotation marks).

utils.h
```diff
@@ -0,0 +1,13 @@
...
+/* TIMER */
+typedef struct {
+    uint64_t begin;
+    uint64_t freq;
+} semu_timer_t;
+
+void semu_timer_init(semu_timer_t *timer, uint64_t freq);
+uint64_t semu_timer_gettime(semu_timer_t *timer);
```
jserv, line 12: Let's unify the naming scheme. That is, consider the following functions:
* `semu_timer_init`
* `semu_timer_get`
* `semu_timer_rebase`

Thus, they all have the consistent prefix and the manner to perform actions.

utils.h
chiangkd: Is it sensible to modify `MaxEmptyLinesToKeep` in `.clang-format` from 3 to 1?

jserv: > Is it sensible to modify `MaxEmptyLinesToKeep` in `.clang-format` from 3 to 1?

No, stick to the current `MaxEmptyLinesToKeep` setting. Sometimes, we do need blank lines. This is particularly true for domain-specific languages with C preprocessor enablement.

chiangkd: Got it, thanks.

jserv:
> SEMU experiences a slowdown as the number of simulated harts increases. This is due to the increased number of timer interrupts and `clock_gettime` system calls in this commit.

Have you tried `CLOCK_REALTIME` as the clock source?

chiangkd:
> Have you tried `CLOCK_REALTIME` as the clock source?

Yes, `CLOCK_REALTIME` perform a similar result for booting system. `CLOCK_REALTIME` is affected by NTP. In the absence of time adjustments, `CLOCK_REALTIME` and `CLOCK_MONOTIC` might show similar time differences.

Execute with `CLOCK_REALTIME`
```
[   32.718809] rcu: INFO: rcu_sched self-detected stall on CPU
...
[   76.194389] Run /init as init process
```

Execute with `CLOCK_MONOTIC`
```
[   32.545280] rcu: INFO: rcu_sched self-detected stall on CPU
...
[   74.768311] Run /init as init process
```

jserv:
> `CLOCK_REALTIME` perform a similar result for booting system. `CLOCK_REALTIME` is affected by NTP. In the absence of time adjustments, `CLOCK_REALTIME` and `CLOCK_MONOTIC` might show similar time differences.

How about `CLOCK_MONOTONIC_RAW` or `CLOCK_MONOTONIC_COARSE`? 
They are implemented as faster but less precise alternatives of `CLOCK_MONOTONIC`.

chiangkd:
> How about `CLOCK_MONOTONIC_RAW` or `CLOCK_MONOTONIC_COARSE`?
> They are implemented as faster but less precise alternatives of `CLOCK_MONOTONIC`.

`CLOCK_MONOTONIC_COARSE` and `CLOCK_REALTIME_COARSE` can be faster because they use less precise and vDSO mechanism. However, they can still trigger RCU warnings.

Execute with `CLOCK_MONOTONIC_COARSE`
```
[   29.840317] rcu: INFO: rcu_sched self-detected stall on CPU
...
[   62.512664] Run /init as init process
...

```
Execute with `CLOCK_REALTIME_COARSE`
```
[   29.979120] rcu: INFO: rcu_sched self-detected stall on CPU
...
[   56.826371] Run /init as init process
```

Use `perf` to monitor performance for different clock sources during 100,000,000 calls to clock_gettime().
|| MONOTONIC | REALTIME | MONOTONIC_COARSE | REALTIME_COARSE |
|-------------| ------------- | ------------- |------------- | ------------- |
|`user`| 1.723s  | 1.796s |0.556s |0.568s|
|`sys`|  0s |  0.008 |0s | 0s|

jserv:
> `CLOCK_MONOTONIC_COARSE` and `CLOCK_REALTIME_COARSE` can be faster because they use less precise and vDSO mechanism. However, they can still trigger RCU warnings.

Based on the experiments, empirical evidence suggests utilizing `CLOCK_MONOTONIC_COARSE`, which is available on modern Linux systems. Please submit a GitHub issue regarding RCU warnings once this pull request is merged.

jserv:
I defer to @ranvd for confirmation.

jserv:

PR #52, opened 2024-08-26
Bump Buildroot version and select GCC 14 with binutils 2.42

shengwen-tw:
This commit intends to avoid a recent glibc bug that mistakenly uses 64-bit time on 32-bit machines.

Consider the Linux `input_event` structure:

```C
struct input_event {
	struct timeval time;
	__u16 type;
	__u16 code;
	__s32 value;
};
```

On RV32 Linux, the total size of this structure should be 16 bytes (8+2+2+4). However, due to a known glibc bug [2] that incorrectly utilizes the 64-bit time type, the actual size of the structure becomes 24 bytes (16+2+2+4).

Without this fix, the bug would affect the usage of `DirectFB2` in the semu project. Fortunately, upgrading the toolchain version resolves the issue.

[1] https://lore.kernel.org/lkml/caa041d27b0fa45aad09a9a262038e3ae4099ca2.camel@mwa.re/T/
[2] https://sourceware.org/pipermail/libc-help/2024-January/006563.html


jserv (changes requested):
Minimize `configs/buildroot.config`. Leave the necessary entries.

jserv:

PR #54, opened 2024-08-31
Enhance precision and safety in time calculations

visitorckw:
Introduces a new inline function, mult_frac(), to handle fractional multiplication with improved precision and safety by computing x * n / d while avoiding common issues like overflow and precision loss. Additionally, the semu_timer_clocksource() function has been refactored to use mult_frac() for time-related calculations, including the calculations for converting time derived from clock_gettime() and mach_absolute_time() into clock ticks. These updates collectively enhance the accuracy and reliability of time computations throughout the codebase.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -19,6 +19,19 @@
 #endif
 #endif
 
+/* Calculate "x * n / d" without unnecessary overflow or loss of precision.
+ *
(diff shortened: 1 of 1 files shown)


jserv:

PR #56, opened 2024-09-03
Rename macros with SEMU_FEATURE prefix

Cuda-Chen:
Rename macros with SEMU_FEATURE prefix
from SEMU_FEATUREXXX to SEMU_FEATURE_XXX.


jserv:
Thank @Cuda-Chen for contributing!

PR #57, opened 2024-09-04
Introduce PACKED macro

shengwen-tw:
Introduce PACKED macro for cross-platform structure alignment, frequently used by VirtIO device implementations in the semu.


jserv:

PR #58, opened 2024-09-09
Fix build on macOS

ChinYikMing:
utils.c:49:9: error: use of undeclared identifier 'mach_clk'


jserv:
Thank @ChinYikMing for contributing! You might consider adding macOS builds to the CI pipelines to ensure that both semu and rv32emu can be built on macOS for major configurations.

PR #59, opened 2024-09-09
CI: Build on macOS

ChinYikMing:


ChinYikMing:
When running boot tests on macOS, the kernel gets stuck at the following message:

`[ 3062.927108] sched_clock: 64 bits at 65MHz, resolution 15ns, wraps every 2199023255550ns`

After bisecting, #49 was identified as the cause of the issue. Cc: @chiangkd

chiangkd:
Hi @ChinYikMing ,

I noticed an abnormal timestamp in the boot log:
> `[ 3062.927108] sched_clock: 64 bits at 65MHz, resolution 15ns, wraps every 2199023255550ns`

This issue arises because I inadvertently returned the clocksource in nanoseconds when building on macOS, while the expected unit is seconds.

```diff
static uint64_t semu_timer_clocksource(uint64_t freq)
{
#if defined(HAVE_POSIX_TIMER)
    struct timespec t;
    clock_gettime(CLOCKID, &t);
    return t.tv_sec * freq + mult_frac(t.tv_nsec, freq, 1e9);
#elif defined(HAVE_MACH_TIMER)
    static mach_timebase_info_data_t t;
    if (t.denom == 0)
        (void) mach_timebase_info(&t);
-    return mult_frac(mach_absolute_time() * freq, t.numer, t.denom);
+    return mult_frac(mult_frac(mach_absolute_time(), freq, 1e9), t.numer, t.denom);
#else
    return time(0) * freq;
#endif
}
```

I will create another PR to fix this bug. Thanks for pointing it out!

Regarding the CI test failures, the CI workflow was failing due to a timeout (currently set to 90 seconds), I found some [discussion](https://github.com/actions/runner-images/issues/1336) about github-hosted runner. When I increasing timeout value to 600 seconds [allowed the workflow to pass](https://github.com/chiangkd/semu/actions/runs/10798598092), I am not sure where is the bottleneck during workflow but it actually take much more time to boot linux on semu with macOS runner. I'm also exploring options to build semu on a real macOS machine to better address this issue.

jserv (changes requested):
Rebase the latest `master` branch.

ChinYikMing:
@chiangkd ，Is the 600s timeout the minimum needed to pass the CI on the macOS runner? I would prefer adding a condition based on the runner to select the appropriate timeout duration. After that, a new issue can be created to track this.

ChinYikMing:
> Rebase the latest `master` branch.

Done.

chiangkd:
> @chiangkd , Is the 600s timeout the minimum needed to pass the CI on the macOS runner? I would prefer adding a conditional statement based on the runner to select the appropriate timeout duration. After that, a new issue can be created to track this.

In my testing, the average time to pass this Action is around 311 seconds. Additionally, I reverted to using the `mtime` counter instead of the real-time clock, which brought the execution time down to about 135 seconds, 

- [Test with realtime clocksource](https://github.com/chiangkd/semu/actions/runs/10798598092)
- [Test with mtime counter](https://github.com/chiangkd/semu/actions/runs/10836556878).

I also tested running semu on macOS via VirtualBox, but it still took longer to boot the system.

Here are the results for `make check` on macOS through VirtualBox:
```shell
...
[142.702404] Run/ init as init process
...
```

.ci/autorun.sh
```diff
@@ -16,8 +16,15 @@ function ASSERT {
 
 cleanup
 
+UNAME_S=$(uname -s)
```
jserv, line 19: Add proper comments to address the difference.

ChinYikMing:
> > @chiangkd , Is the 600s timeout the minimum needed to pass the CI on the macOS runner? I would prefer adding a conditional statement based on the runner to select the appropriate timeout duration. After that, a new issue can be created to track this.
> 
> In my testing, the average time to pass this Action is around 311 seconds. Additionally, I reverted to using the `mtime` counter instead of the real-time clock, which brought the execution time down to about 135 seconds,
> 
> * [Test with realtime clocksource](https://github.com/chiangkd/semu/actions/runs/10798598092)
> * [Test with mtime counter](https://github.com/chiangkd/semu/actions/runs/10836556878)
> 
> I also tested running semu on macOS via VirtualBox, but it still took longer to boot the system.
> 
> Here are the results for `make check` on macOS through VirtualBox:
> 
> ```shell
> ...
> [142.702404] Run/ init as init process
> ...
> ```

Got it, thanks for testing. It looks like 600 seconds is sufficient to pass the CI on the macOS runner.

jserv:

PR #60, opened 2024-09-10
Fix incorrect clock calculation on macOS

chiangkd:
The current `semu_timer_clocksource()` returns the clock source in nanoseconds when building on macOS. Causing an abnormal timestamp in the boot log.
```
...
[ 3062.927108] sched_clock: 64 bits at 65MHz, resolution 15ns, wraps every 2199023255550ns
```
Divide the return value by 1e9, ensuring the clock source is provided in seconds.

### DIFF
--- a/utils.c
+++ b/utils.c
@@ -48,7 +48,8 @@ static uint64_t semu_timer_clocksource(uint64_t freq)
     static mach_timebase_info_data_t t;
     if (t.denom == 0)
         (void) mach_timebase_info(&t);
-    return mult_frac(mach_absolute_time() * freq, t.numer, t.denom);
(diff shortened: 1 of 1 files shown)


ChinYikMing:
Change `MacOS` in the commit title to `macOS`.

ChinYikMing:
Issue found in #59.

ChinYikMing:
> Change `MacOS` in the commit title to `macOS`.

Also the title in the commit, you may force push after changing.

jserv (changes requested):
Improve the git commit message.

jserv:

PR #61, opened 2024-09-12
Optimize semu_timer_clocksource() precision on macOS

visitorckw:
Improve the precision of semu_timer_clocksource() by deferring the division by 1e9 until the final step. Since t.numer and t.denom are typically much smaller compared to the clock frequency and 1e9, performing the division earlier in the calculation could result in a significant loss of precision. This change ensures that precision is preserved by delaying the division until the end of the computation.


ChinYikMing (approved):

chiangkd (approved):

jserv:

PR #64, opened 2024-11-11
Consolidate and extend network client support

chiangkd:
This PR refactors the existing TAP-based network device implementation without affecting the current functionality. It separates network client handling to support multiple backends, allowing the use of the `-netdev` flag to specify the client type, defaulting to TAP if not provided. Support for other backends is yet to be implemented.

Additionally, packet transfer functions are now designed to be specific to each network client, making the system more modular and extensible for future development.

Verification of existing functionality (in privileged mode):
1. Open TAP device and run simulator
```
sudo make check
```
2. Configure the IP address for `tap0` and enable it on the host side:
```
sudo ip addr add 192.168.10.1/24 dev tap0
sudo ip link set tap0 up
```
3. Enable `eth0` and set its IP address:
```
ip l set eth0 up
ip a add 192.168.10.2/24 dev eth0
```


net_client.h
```diff
@@ -0,0 +1,29 @@
...
+typedef struct {
+    char *name;
+    net_client_driver_t type;
+    union {
+        net_tap_options tap;
+        net_user_options user;
+    } op;
+} netdev_t;
+
+int net_client_init(netdev_t *nedtev, const char *net_type);
```
jserv, line 29: Why did you name it `net_client`? Is there explicitly a net "server"?

net_client.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "riscv.h"
```
jserv, line 3: Is the header inclusion necessary?

net_client.h
chiangkd: The choice of `net_client` as a name follows the convention used in QEMU, where `NetClientState` abstracts various network interfaces. In this context, “client” doesn’t imply a traditional client-server relationship but rather represents an individual network endpoint.

`net_backend` might indeed be a clearer choice, as it more directly reflects the structure's role in handling network backends, without implying a client-server relationship. I'll consider updating the name to `net_backend` for improved clarity and consistency.

net_client.h
chiangkd: No, I will remove it. Thank you for pointing it out!

net_client.h
jserv: How about `netdev` which represent network device?

chiangkd:
> Enhance CI pipeline to automate the network specific tests.
> 
> See [vwifi](https://github.com/sysprog21/vwifi) for reference.

netdev.c
```diff
@@ -0,0 +1,80 @@
...
+
+    fprintf(stderr, "allocated TAP interface: %s\n", ifreq.ifr_name);
+    assert(fcntl(tap->tap_fd, F_SETFL,
+                 fcntl(tap->tap_fd, F_GETFL, 0) | O_NONBLOCK) >= 0);
+    return 0;
+}
+
+static int net_init_slirp(netdev_t *netdev)
+{
+    // TBD: create slirp dev
```
jserv, line 65: Use `TODO` and C-style comments.

netdev.c
```diff
@@ -0,0 +1,80 @@
...
+    return 0;
+}
+
+int netdev_init(netdev_t *netdev, const char *net_type)
+{
+    int dev_idx = find_net_dev_idx(net_type, netdev_driver_lookup);
+    if (dev_idx == -1)
+        return false;
+    else
```
jserv, line 74: Remove `else` statement here.

virtio-net.c
```diff
@@ -223,15 +273,25 @@ void virtio_net_refresh_queue(virtio_net_state_t *vnet)
...
-    if (pfd.revents & POLLIN) {
-        vnet->queues[VNET_QUEUE_RX].fd_ready = true;
-        virtio_net_try_rx(vnet);
-    }
-    if (pfd.revents & POLLOUT) {
-        vnet->queues[VNET_QUEUE_TX].fd_ready = true;
-        virtio_net_try_tx(vnet);
+    net_dev_driver_t dev_type = vnet->peer.type;
+    switch (dev_type) {
+    case NET_DEV_DRIVER_TAP:
```
jserv, line 278: Add `{` and `}` around `case NET_DEV_DRIVER_TAP:`

netdev.h
```diff
@@ -0,0 +1,27 @@
...
+    NET_DEV_DRIVER_MAX,
+} net_dev_driver_t;
+
+typedef struct {
+    int tap_fd;
+} net_tap_options;
+
+typedef struct {
+    /* TODO: Implement user option */
+    int temp;
```
jserv, line 15: Don't fill the structure since it is unimplemented.

netdev.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+typedef enum {
+    NET_DEV_DRIVER_TAP,
+    NET_DEV_DRIVER_USER,
+    NET_DEV_DRIVER_MAX,
+} net_dev_driver_t;
+
+typedef struct {
```
jserv, line 9: Can you hide the implementation details as possible?

chiangkd: I use an X-Macro to define supported network devices:
```c
#define SUPPORTED_DEVICES   \
        _(tap)              \
        _(user)
```
Previously, `NET_DEV_DRIVER_MAX` was used to denote the total number of network device drivers implemented. To simplify the design, this constant has been removed. Instead, the find_net_dev_idx function dynamically determines the range of the lookup table based on its content.

The final `NULL` ensures the array is properly terminated, allowing for safe iteration.
```c
static const char *netdev_driver_lookup[] = {
#define _(dev) [NET_DEV_DRIVER_##dev] = #dev,
    SUPPORTED_DEVICES
#undef _
        NULL,
};

...

static int find_net_dev_idx(const char *net_type, const char **netlookup)
{
    if (!net_type)
        return -1;
    int i = 0;
    while (netlookup[i]) {
        if (!strcmp(net_type, netlookup[i])) {
            return i;
        }
        i++;
    }
    return -1;
}

```

netdev.c
```diff
@@ -0,0 +1,87 @@
...
+{
+    if (!net_type)
+        return -1;
+    int i = 0;
+    while (netlookup[i]) {
+        if (!strcmp(net_type, netlookup[i])) {
+            return i;
+        }
+        i++;
+    }
```
jserv, line 33: Replace the code snip with for-iteration.

netdev.c
```diff
@@ -0,0 +1,87 @@
...
+    return 0;
+}
+
+static int net_init_user(netdev_t *netdev)
+{
+    /* TODO: create slirp dev */
+    return 0;
+}
+
+int netdev_init(netdev_t *netdev, const char *net_type)
```
jserv, line 67: Refine the return type. Should be `bool`.

netdev.c
```diff
@@ -0,0 +1,84 @@
...
+        break;
+        SUPPORTED_DEVICES
+#undef _
+    default:
+        fprintf(stderr, "unknown network device\n");
+        break;
+    }
+
+    return true;
+}
```
jserv, line 84: Do [add a new line to the end of a file](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file).

netdev.h
```diff
@@ -0,0 +1,34 @@
...
+    /* TODO: Implement user option */
+} net_user_options;
+
+typedef struct {
+    char *name;
+    net_dev_driver_t type;
+    union {
+        net_tap_options tap;
+        net_user_options user;
+    } op;
```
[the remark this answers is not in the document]
chiangkd: I try to encapsulate the implementation details of `netdev_t` as follow:
```c
typedef struct {
    char *name;
    net_dev_driver_t type;
    void *device_op;
} netdev_t;
```
The `device_op` pointer is allocated and defined during the `netdev_init` process, which is called by `virtio_net_init`. Also read/write operations in `virtio-net` depend on the underlying peer device specified by `netdev_t`.
```c
static ssize_t handle_write(netdev_t *netdev,
                            virtio_net_queue_t *queue,
                            struct iovec *iovs_cursor,
                            size_t niovs)
{
    ssize_t plen = 0;
#define _(dev) NET_DEV_DRIVER_##dev
    switch (netdev->type) {
    case _(tap):
        net_tap_options_t *tap = (net_tap_options_t *) netdev->device_op;
        plen = writev(tap->tap_fd, iovs_cursor, niovs);
        if (plen < 0 && (errno == EWOULDBLOCK || errno == EAGAIN)) {
            queue->fd_ready = false;
            return -1;
        }
        if (plen < 0) {
            plen = 0;
            fprintf(stderr, "[VNET] could not write packet: %s\n",
                    strerror(errno));
        }
        break;
        ...
    }
#undef _
    return plen;
}
```

netdev.h
```diff
@@ -0,0 +1,31 @@
...
+} net_tap_options_t;
+
+typedef struct {
+    /* TODO: Implement user option */
+} net_user_options_t;
+
+typedef struct {
+    char *name;
+    net_dev_driver_t type;
+    void *device_op;
```
jserv, line 28: Simply name it `op`, which stands for "operation."

netdev.h
```diff
@@ -0,0 +1,31 @@
...
+#define SUPPORTED_DEVICES   \
+        _(tap)              \
+        _(user)
+/* clang-format on */
+
+typedef enum {
+#define _(dev) NET_DEV_DRIVER_##dev,
+    SUPPORTED_DEVICES
+#undef _
+} net_dev_driver_t;
```
jserv, line 15: Rename it to `netdev_impl_t` for consistency.

netdev.c
```diff
@@ -0,0 +1,90 @@
...
+#include <string.h>
+#include <sys/ioctl.h>
+
+#include "netdev.h"
+
+static int net_init_tap();
+static int net_init_user();
+
+static const char *netdev_driver_lookup[] = {
+#define _(dev) [NET_DEV_DRIVER_##dev] = #dev,
```
jserv, line 17: Enforce the consistent prefix `NETDEV_`. That is, use `NETDEV_IMPL_` here.

jserv:

PR #65, opened 2024-12-10
Enable ISIG control mode of termios after exit

ChinYikMing:
How to reproduce?
1. Run the emulation.
```
$ make check
```

2. Terminate the emulation via keyboard `CTRL + a + x`.

3. Run some shell interactive program (e.g., cat).
```
$ cat
```

4. Try to terminate the shell interactive program (e.g., cat) via keyboard `CTRL + c`.
The SIGINT cannot be caught and propagate to the `cat` process. Example output:
```
$ cat
apple
apple
^C^C^C^C^C
```

This patch fix this issue.


uart.c
```diff
@@ -10,6 +10,8 @@
 #include "riscv.h"
 #include "riscv_private.h"
 
+#define TERMIOS_C_CFLAG (ICANON | ECHO | ISIG)
```
jserv, line 13: Add comments for the configuration.

jserv:

PR #67, opened 2025-01-11
Scale frequency to suppress RCU CPU stall warning

Mes0903:
Since the emulator currently operates using sequential emulation, the execution time for the boot process is relatively long, which can result in the generation of RCU CPU stall warnings.

To address this issue, there are several potential solutions:

1. Scale the frequency to slow down emulator time during the boot process, thereby eliminating RCU CPU stall warnings.
2. During the boot process, avoid using `clock_gettime` to update ticks and instead manage the tick increment relationship manually.
3. Implement multi-threaded emulation to accelerate the emulator's execution speed.

For the third point, while implementing multi-threaded emulation can significantly accelerate the emulator's execution speed, it cannot guarantee that this issue will not reappear as the number of cores increases in the future. Therefore, a better approach is to use methods 1 and 2 to allow the emulator to set an expected time for completing the boot process.

The advantages and disadvantages of the scale method are as follows:

Advantages:
- Simple implementation
- Effectively sets the expected boot process completion time
- Results have strong interpretability
- Emulator time can be easily mapped back to real time

Disadvantages:
- Slower execution speed

The advantages and disadvantages of the increment ticks method are as follows:

Advantages:
- Faster execution speed
- Effectively sets the expected boot process completion time

Disadvantages:
- More complex implementation
- Some results are difficult to interpret
- Emulator time is difficult to map back to real time

Based on practical tests, the second method provides limited acceleration but introduces some significant drawbacks, such as difficulty in interpreting results and the complexity of managing the increment relationship. Therefore, this commit opts for the scale frequency method to address this issue.

This commit divides time into emulator time and real time. During the boot process, the timer uses scale frequency to slow down the growth of emulator time, eliminating RCU CPU stall warnings. After the boot process is complete, the growth of emulator time aligns with real time.

To configure the scale frequency parameter, three pieces of information are required:

1. The expected completion time of the boot process
2. A reference point for estimating the boot process completion time
3. The relationship between the reference point and the number of SMPs

According to the Linux kernel documentation:
https://docs.kernel.org/RCU/stallwarn.html#config-rcu-cpu-stall-timeout

The grace period for RCU CPU stalls is typically set to 21 seconds. By dividing this value by two as the expected completion time, we can provide a sufficient buffer to reduce the impact of errors and avoid RCU CPU stall warnings.

Using gprof for basic statistical analysis, it was found that `semu_timer_clocksource` accounts for approximately 10% of the boot process execution time. Since the logic within `semu_timer_clocksource` is relatively simple, its execution time can be assumed to be nearly equal to `clock_gettime`.

Furthermore, by adding a counter to `semu_timer_clocksource`, it was observed that each time the number of SMPs increases by 1, the execution count of `semu_timer_clocksource` increases by approximately $2 \times 10^8$ (see the table below).

With this information, we can estimate the boot process completion time as 

$$
\text{SecPerCall} \times \text{SMPs} \times 2 \times 10^8 \times \frac{100\%}{10\%}
$$

seconds, and thereby calculate the scale frequency parameter. For instance, if the estimated time is 200 seconds and the target time is 10 seconds, the scaling factor would be `10 / 200`.

To calculate the proportion of `semu_timer_clocksource` using gprof, simply add the `-pg` option during compilation. Then, terminate the emulator's operation immediately after the first switch to U mode (when the boot process is complete). This approach allows for a rough estimation of the proportion occupied by `semu_timer_clocksource`.

Below is the gprof testing output:

- CPU: 13th Gen Intel(R) Core(TM) i7-13700
	```
	mes@DESKTOP-HLQ9F6A:~/semu$ gprof ./semu
	Flat profile:
	
	Each sample counts as 0.01 seconds.
	  %   cumulative   self              self     total           
	 time   seconds   seconds    calls  Ts/call  Ts/call  name    
	 39.91     55.19    55.19                             vm_step
	 10.55     69.78    14.59                             semu_timer_clocksource
	  9.27     82.60    12.82                             semu_start
	  6.49     91.57     8.97                             aclint_mswi_update_interrupts
	  6.19    100.13     8.56                             mmu_store
	  5.73    108.05     7.92                             mmu_translate
	  3.98    113.56     5.51                             semu_timer_get
	  3.33    118.16     4.60                             op_rv32i
	  2.92    122.20     4.04                             aclint_mtimer_update_interrupts
	  2.76    126.02     3.82                             aclint_sswi_update_interrupts
	  2.10    128.93     2.91                             ram_read
	  1.59    131.13     2.20                             mmu_load
	  1.27    132.89     1.76                             mem_store
	  1.19    134.54     1.65                             _init
	  0.62    135.40     0.86                             mem_load
	  0.46    136.04     0.64                             u8250_check_ready
	  0.41    136.61     0.57                             ram_write
	  0.39    137.15     0.54                             csr_read
	  0.37    137.66     0.51                             virtio_net_refresh_queue
	  0.28    138.05     0.39                             mem_fetch
	  0.17    138.28     0.23                             op_csr_cs
	```

- CPU: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
	```
	mes@Mes:~/semu$ gprof ./semu
	Flat profile:
	
	Each sample counts as 0.01 seconds.
	  %   cumulative   self              self     total           
	 time   seconds   seconds    calls  Ts/call  Ts/call  name    
	 32.96     77.77    77.77                             vm_step
	 15.96    115.43    37.66                             semu_timer_clocksource
	 13.95    148.34    32.91                             semu_start
	  6.10    162.74    14.40                             aclint_mswi_update_interrupts
	  5.95    176.77    14.03                             mmu_store
	  5.35    189.39    12.62                             mmu_translate
	  3.85    198.47     9.08                             aclint_mtimer_update_interrupts
	  2.73    204.91     6.44                             op_rv32i
	  2.46    210.72     5.81                             ram_read
	  2.11    215.70     4.98                             aclint_sswi_update_interrupts
	  1.74    219.81     4.11                             mmu_load
	  1.74    223.91     4.10                             semu_timer_get
	  1.30    226.98     3.07                             mem_store
	  0.92    229.14     2.16                             ram_write
	  0.81    231.05     1.91                             _init
	  0.61    232.48     1.43                             mem_load
	  0.55    233.78     1.30                             csr_read
	  0.35    234.60     0.82                             u8250_check_ready
	  0.22    235.11     0.51                             op_csr_cs
	  0.21    235.61     0.50                             virtio_net_refresh_queue
	  0.14    235.95     0.34                             mem_fetch
	  0.00    235.96     0.01                             semu_timer_init
	```

And by adding a counter to `semu_timer_clocksource`, it becomes possible to calculate the relationship between SMPs and the number of times `semu_timer_clocksource` is called.

Below is the testing output(13th Gen Intel(R) Core(TM) i7-13700):

| SMP | times call `semu_timer_total_ticks` |
| --- | ----------------------------------- |
| 1   | 239,937,385                         |
| 2   | 410,377,969                         |
| 3   | 600,190,253                         |
| 4   | 825,078,230                         |
| 5   | 1,007,098,511                       |
| 6   | 1,213,304,632                       |
| 7   | 1,419,857,500                       |
| 8   | 1,627,353,803                       |
| 9   | 1,835,991,887                       |
| 10  | 2,056,837,458                       |
| 11  | 2,269,651,319                       |
| 12  | 2,485,299,111                       |
| 13  | 2,703,809,852                       |
| 14  | 2,917,880,774                       |
| 15  | 3,134,960,385                       |
| 16  | 3,450,041,050                       |
| 17  | 3,587,880,492                       |
| 18  | 3,810,315,132                       |
| 19  | 4,052,935,937                       |
| 20  | 4,274,307,156                       |
| 21  | 4,503,423,577                       |
| 22  | 4,742,850,891                       |
| 23  | 4,973,145,039                       |
| 24  | 5,213,953,075                       |
| 25  | 5,597,276,263                       |
| 26  | 5,696,829,697                       |
| 27  | 5,935,553,504                       |
| 28  | 6,173,571,960                       |
| 29  | 6,620,649,951                       |
| 30  | 6,683,572,727                       |
| 31  | 6,942,616,048                       |
| 32  | 7,180,650,711                       | 
 <div id='description'>
<h3>Summary by Bito</h3>
Implementation of frequency scaling solution to address RCU CPU stall warnings during boot process. The solution divides time handling between emulator and real time, using scaled frequency during boot and switching to real-time after completion. Targets 10-second boot time with comprehensive testing infrastructure across SMP configurations.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>


jserv (changes requested):
Don't create a file named `timer.h` which causes the confusion. The proposed timer.[ch] has no direct relationship with hardware timer. Thus, apply the proposed changes to `utils.c`.

utils.c
```diff
@@ -19,48 +19,160 @@
...
 
+#ifndef SEMU_SMP
+#define SEMU_SMP 1
+#endif
+
+#ifndef SEMU_BOOT_TARGET_TIME
+#define SEMU_BOOT_TARGET_TIME 10
+#endif
+
+bool boot_complete = false;
```
ranvd, line 30: I suggest moving this variable into `vm_t` for a more conceptually accurate design.

Mes0903: If we move `boot_complete` into `vm_t`, all existing functions would need an additional `vm_t` parameter. For example, `semu_timer_get` would change to:

```c
semu_timer_get(vm_t *vm, semu_timer_t *timer)
```

This change would indirectly require the areas that call this function to also pass in a `vm_t`parameter. For instance, since `semu_timer_get` is called within `aclint_mtimer_update_interrupts`, the API of `aclint_mtimer_update_interrupts` would also need to be updated to include `vm_t`.

As this pattern continues, the API changes would proliferate significantly. Perhaps we could introduce a static bool pointer pointing to `boot_complete` and assign its value during `semu_timer_init`. This way, we would only need to modify the parameters of `semu_timer_init`.

utils.c
```diff
@@ -19,48 +19,160 @@
...
+/* Measure how long a single 'clock_gettime' takes, to scale real time in order
+ * to set the emulator time.
+ */
+static void measure_bogomips_ns(uint64_t target_loop)
+{
+    struct timespec start, end;
+    clock_gettime(CLOCKID, &start);
+
+    for (uint64_t loops = 0; loops < target_loop; loops++)
+        clock_gettime(CLOCKID, &end);
```
jserv, line 68: I did not get the idea behind this code snip. What did you update the value `end` several times?

utils.c
```diff
@@ -19,48 +19,160 @@
...
+ *
+ * Thus, by seconds * frequency + nanoseconds * frequency / 1,000,000,000, we
+ * can get the number of ticks.
+ */
+static inline uint64_t get_ticks(struct timespec *ts, double freq)
+{
+    return ts->tv_sec * freq + mult_frac(ts->tv_nsec, freq, 1000000000ULL);
+}
+
+/* Measure how long a single 'clock_gettime' takes, to scale real time in order
```
jserv, line 59: Replace `clock_gettime` with high resolution timer.

utils.c
```diff
@@ -19,48 +19,160 @@
...
 #if defined(HAVE_POSIX_TIMER)
-    struct timespec t;
-    clock_gettime(CLOCKID, &t);
-    return t.tv_sec * freq + mult_frac(t.tv_nsec, freq, 1e9);
+    struct timespec emulator_time;
+    clock_gettime(CLOCKID, &emulator_time);
+
+    if (!boot_complete) {
+        return get_ticks(&emulator_time, timer->freq * scale_factor);
+    } else {
```
jserv, line 114: Remove the `else` statement due to the `return` state right before this line.

utils.c
Mes0903: Here is meant to measure the execution time of a single `clock_gettime` call. In my understanding, the following code can achieve the same purpose:

```c
struct timespec start, end;
clock_gettime(CLOCKID, &start);

for (uint64_t loops = 0; loops < target_loop - 1; loops++) {
    struct timespec ts;
    clock_gettime(CLOCKID, &ts);
}

clock_gettime(CLOCKID, &end);
```

However, in the for loop, the variable `ts` is not used. Therefore, I simply replaced it with `end` as the parameter of `clock_gettime`. This way, when exiting the loop, the value of `end` will be the time from the last execution of `clock_gettime`, which can also achieve the purpose of measure the execution time of a single `clock_gettime`.

jserv: You should add comments to address the measurement for HRT syscall overhead.

jserv: By the way, you should wrap POSIX, macOS, and time utility in standard library.

Mes0903:
I noticed that the time duting boot process was far from the target time on macOS. I think this may cause by the hypothesis of SMP*2*10^8 calls of `semu_timer_clocksource` or `semu_timer_clocksource` is ~10% of boot overhead, is not correct on macOS.

utils.c
```diff
@@ -19,48 +19,189 @@
...
+ *    seconds.
+ */
+static void measure_bogomips_ns(uint64_t target_loop)
+{
+    /* Mark start time in ns */
+    uint64_t start_ns = host_time_ns();
+
+    /* Perform 'target_loop' times calling the host HRT. */
+    for (uint64_t loops = 0; loops < target_loop; loops++)
+        (void) host_time_ns();
```
jserv, line 105: See [Accurate Benchmarking](https://ates.dev/posts/2025-01-12-accurate-benchmarking/) to eliminate loop overhead.

utils.h
```diff
@@ -8,6 +9,8 @@ typedef struct {
     uint64_t freq;
 } semu_timer_t;
 
+extern bool boot_complete; /* complete boot process and get in initrd */
```
jserv, line 12: Refine the definition: time to reach the first process.

utils.c
```diff
@@ -19,48 +19,189 @@
...
-    if (t.denom == 0)
-        (void) mach_timebase_info(&t);
-    return mult_frac(mult_frac(mach_absolute_time(), t.numer, t.denom), freq,
-                     1e9);
+    static mach_timebase_info_data_t ts = {0};
+    if (ts.denom == 0)
+        (void) mach_timebase_info(&ts);
+
+    uint64_t now = mach_absolute_time();
+    // convert to nanoseconds: (now * t.numer / t.denom)
```
jserv, line 78: Always use C-style comments.

utils.c
```diff
@@ -19,48 +19,189 @@
...
+    return mult_frac(now, (double) ts.numer, (uint64_t) ts.denom);
+
 #else
-    return time(0) * freq;
+    /* Minimal fallback: time(0) in seconds => convert to ns. */
+    time_t now_sec = time(0);
+    return (uint64_t) now_sec * 1000000000ULL;
```
jserv, line 84: Replace the numerics with `1e9`.

utils.c
```diff
@@ -19,48 +19,189 @@
 #endif
 #endif
 
+#ifndef SEMU_SMP
+#define SEMU_SMP 1
+#endif
```
jserv, line 24: Don't do this. The macro `SEMU_SMP` should be set properly by build system.

utils.c
```diff
@@ -19,48 +19,203 @@
...
+bool boot_complete = false;
+static double scale_factor;
+
 /* Calculate "x * n / d" without unnecessary overflow or loss of precision.
  *
  * Reference:
  * https://elixir.bootlin.com/linux/v6.10.7/source/include/linux/math.h#L121
  */
-static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
+static inline uint64_t mult_frac(uint64_t x, double n, uint64_t d)
```
jserv, line 30: Don't change the function prototype. Instead, scale the floating-point values for integer-only arithmetics.

utils.h
```diff
@@ -8,6 +9,8 @@ typedef struct {
     uint64_t freq;
 } semu_timer_t;
 
+extern bool boot_complete; /* Time to reach the first user process. */
+
 void semu_timer_init(semu_timer_t *timer, uint64_t freq);
 uint64_t semu_timer_get(semu_timer_t *timer);
-void semu_timer_rebase(semu_timer_t *timer, uint64_t time);
\ No newline at end of file
+void semu_timer_rebase(semu_timer_t *timer, uint64_t time);
```
jserv, line 16: Don't change this line.
See https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e

jserv:
Test under Apple M1:
```
$ make SMP=4 check
...
[  388.139490] rcu: INFO: rcu_sched self-detected stall on CPU
[  388.156300] rcu: 	0-....: (41291 ticks this GP) idle=bdfc/1/0x40000002 softirq=1137/1137 fqs=6808
[  388.187007] 	(t=68790 jiffies g=-327 q=5 ncpus=4)
[  388.205900] CPU: 0 PID: 37 Comm: kworker/u8:1 Not tainted 6.1.99 #1
[  388.226518] Hardware name: semu (DT)
[  388.239863] Workqueue: events_unbound async_run_entry_fn
[  388.260077] epc : do_reset+0x30/0x6c
[  388.278895]  ra : do_reset+0x38/0x6c
[  388.297349] epc : c03816d0 ra : c03816d8 sp : c09d7e50
[  388.314871]  gp : c04de840 tp : c09d8000 t0 : c0bd3e50
[  388.332163]  t1 : 00000008 t2 : 8a9164d2 s0 : c09d7e60
[  388.349003]  s1 : c03a48b8 a0 : 00000001 a1 : 00000000
[  388.365395]  a2 : 00000000 a3 : 007a5804 a4 : 0005a7fb
[  388.381809]  a5 : 00000000 a6 : 00000000 a7 : 0000000c
[  388.398328]  s2 : 004725a7 s3 : 0038da58 s4 : 00000000
[  388.415347]  s5 : 00000030 s6 : 00000007 s7 : c03a496c
[  388.432367]  s8 : c03a48b8 s9 : c0381488 s10: c03a4968
[  388.450494]  s11: c0381894 t3 : c0b9d440 t4 : 00000003
[  388.467256]  t5 : 94d7d78e t6 : c09d7e08
[  388.481446] status: 00000120 badaddr: dfea5804 cause: 80000005
[  388.500241] [<c03816d0>] do_reset+0x30/0x6c
[  388.521698] [<c0381870>] write_buffer+0x40/0x64
[  388.543520] [<c038207c>] unpack_to_rootfs+0x1c4/0x300
[  388.579518] [<c0382848>] do_populate_rootfs+0x70/0xd8
[  388.601539] [<c002bd8c>] async_run_entry_fn+0x30/0xd0
[  388.623279] [<c0021990>] process_one_work+0x198/0x224
[  388.644797] [<c0021f18>] worker_thread+0x238/0x294
[  388.667502] [<c00289c0>] kthread+0xd4/0xd8
[  388.687176] [<c0002514>] ret_from_exception+0x0/0x1c
[  416.399001] Freeing initrd memory: 8188K
...
```

jserv:
Test on Ubuntu Linux + [eMAG](https://www.anandtech.com/show/15733/ampere-emag-system-a-32core-arm64-workstation/4) (32-core Armv8-A):
```
$ make SMP=32 check
...
[   35.995833] rcu: INFO: rcu_sched self-detected stall on CPU
[   36.000832] rcu: 	6-....: (5249 ticks this GP) idle=111c/1/0x40000002 softirq=1136/1136 fqs=1950
[   36.011131] 	(t=5252 jiffies g=-987 q=11 ncpus=32)
[   36.016552] CPU: 6 PID: 8 Comm: kworker/u64:0 Not tainted 6.1.99 #1
[   36.023057] Hardware name: semu (DT)
[   36.026911] Workqueue: events_unbound async_run_entry_fn
[   36.033356] epc : eat+0x20/0x54
[   36.039018]  ra : do_reset+0x38/0x6c
[   36.045041] epc : c0381544 ra : c03816d8 sp : c0847e40
[   36.050702]  gp : c04de840 tp : c087af40 t0 : c1325e50
[   36.056304]  t1 : 00000008 t2 : e893c869 s0 : c0847e50
[   36.061664]  s1 : c03a48b8 a0 : 00000001 a1 : 00000000
[   36.066844]  a2 : 00000000 a3 : 0054cb37 a4 : dfc4cb38
[   36.072265]  a5 : c03a48b8 a6 : 00000000 a7 : 0000000c
[   36.077565]  s2 : 004725a7 s3 : 0038da58 s4 : 00000000
[   36.083106]  s5 : 00000030 s6 : 00000007 s7 : c03a496c
[   36.088466]  s8 : c03a48b8 s9 : c0381488 s10: c03a4968
[   36.094128]  s11: c0381894 t3 : c130ff40 t4 : 00000003
[   36.099609]  t5 : 56e350ad t6 : c0847e08
[   36.104066] status: 00000120 badaddr: c03a48c4 cause: 80000005
[   36.110149] [<c0381544>] eat+0x20/0x54
[   36.116774] [<c03816d8>] do_reset+0x38/0x6c
[   36.123580] [<c0381870>] write_buffer+0x40/0x64
[   36.130567] [<c038207c>] unpack_to_rootfs+0x1c4/0x300
[   36.138036] [<c0382848>] do_populate_rootfs+0x70/0xd8
[   36.144179] [<c002bd8c>] async_run_entry_fn+0x30/0xd0
[   36.150865] [<c0021990>] process_one_work+0x198/0x224
[   36.157731] [<c0021f18>] worker_thread+0x238/0x294
[   36.164296] [<c00289c0>] kthread+0xd4/0xd8
[   36.169957] [<c0002514>] ret_from_exception+0x0/0x1c
[   70.530512] Freeing initrd memory: 8188K
```

Mes0903:
Based on the output, I think it clearly that the assumption of " $SMP \times 2 \times 10^8$ calls of `semu_timer_clocksource`" or "`semu_timer_clocksource` is ~10% of boot overhead" does not hold on macOS.  We need to re-evaluate the number of `semu_timer_clocksource` calls and its proportion on macOS to accurately calculate the scale factor.

Makefile
```diff
@@ -56,6 +56,8 @@ OBJS := \
 	aclint.o \
 	$(OBJS_EXTRA)
 
+LDFLAGS := -pg
```
jserv, line 59: The [`-pg` instrumentation option](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) might be too heavy. You should consider more lightweight ways.

jserv (changes requested):
Refine the git commit messages.

Change from
> According to the Linux kernel documentation:
https://docs.kernel.org/RCU/stallwarn.html#config-rcu-cpu-stall-timeout

to
> According to Using RCU’s CPU Stall Detector [1]
> ...
> [1] https://docs.kernel.org/RCU/stallwarn.html#config-rcu-cpu-stall-timeout

Append `Close #51` at the end of git commit messages.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+    static mach_timebase_info_data_t ts = {0};
+    if (ts.denom == 0)
+        (void) mach_timebase_info(&ts);
+
+    uint64_t now = mach_absolute_time();
+    /* convert to nanoseconds: (now * t.numer / t.denom) */
+    return mult_frac(now, ts.numer, (uint64_t) ts.denom);
+
+#else
+    /* Minimal fallback: time(0) in seconds => convert to ns. */
```
jserv, line 62: Replace "Minimal fallback" with "Fallback to non-HRT calls"

utils.h
```diff
@@ -1,13 +1,16 @@
...
+extern bool boot_complete; /* Time to reach the first user process. */
+
 /* TIMER */
 typedef struct {
     uint64_t begin;
     uint64_t freq;
 } semu_timer_t;
 
-void semu_timer_init(semu_timer_t *timer, uint64_t freq);
+void semu_timer_init(semu_timer_t *timer, uint64_t freq, int hart_count);
```
jserv, line 14: Replace `hart_count` with "n_harts`.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+        offset = (int64_t) (real_val - scaled_val);
+    }
+
+    /* Return real freq minus offset. */
+    uint64_t real_freq_val = (uint64_t) now_sec * (uint64_t) timer->freq;
+    return real_freq_val - offset;
 #endif
 }
 
+void semu_timer_init(semu_timer_t *timer, uint64_t freq, int hart_count)
```
jserv, line 192: Ditto. Use `n_harts`.

jserv: The "count" is used for iterations.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
     return q * n + r * n / d;
 }
 
-void semu_timer_init(semu_timer_t *timer, uint64_t freq)
+/* On POSIX => use clock_gettime().
+ * On macOS => use mach_absolute_time().
+ * Else => fallback to time(0) in seconds, convert to ns.
```
jserv, line 40: Refine as following:
```
High-precision time measurement:
- POSIX systems: clock_gettime() for nanosecond precision
- macOS: mach_absolute_time() with timebase conversion
- Other platforms: time(0) with conversion to nanoseconds as fallback
```

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+ *
+ * Now, the POSIX/macOS logic can be clearly reused. Meanwhile, the fallback
+ * path might just do a coarser approach with time(0).
```
jserv, line 43: Refine as following:
> The platform-specific timing logic is now clearly separated: POSIX and macOS implementations provide high-precision measurements, while the fallback path uses time(0) for a coarser but portable approach.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+     * iteration, while in the second loop, it is called twice per iteration.
+     *
+     * In this way, the cost of the first loop is 'e + t', and the cost of the
+     * second loop is 'e + 2t'. By subtracting the two, we can effectively
+     * eliminate the loop overhead.
+     *
+     * Reference:
+     * https://ates.dev/posts/2025-01-12-accurate-benchmarking/
+     */
+    const uint64_t start_ns_1 = host_time_ns();
```
jserv, line 95: Rename to `start1_ns`. Make the suffix `_ns` stand for nanoseconds.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+     * minimum here to get more fault tolerance. Thus, effectively:
+     *   predict_sec = ns_per_call * SMP * 2e8 * (100%/10%) / 1e9
+     *               = ns_per_call * SMP * 2.0
+     * Then scale_factor = (desired_time) / (predict_sec).
+     */
+    const double predict_sec = ns_per_call * hart_count * 2.0;
+    scale_factor = SEMU_BOOT_TARGET_TIME / predict_sec;
+}
+
+/* The main function that returns the "emulated time" in ticks.
```
jserv, line 128: Remove "main function" to avoid confusion.

utils.c
```diff
@@ -32,35 +35,177 @@ static inline uint64_t mult_frac(uint64_t x, uint64_t n, uint64_t d)
...
+ * 'clock_gettime()' on POSIX or 'mach_absolute_time()' on macOS.
+ *
+ * 1) Times how long it takes to call 'host_time_ns()' repeatedly (iterations).
+ * 2) Derives an average overhead per call => ns_per_call.
+ * 3) Because semu_timer_clocksource is ~10% of boot overhead, and called ~2e8
+ *    times * SMP, we get predict_sec = ns_per_call * SMP * 2. Then set
+ *    'scale_factor' so the entire boot completes in SEMU_BOOT_TARGET_TIME
+ *    seconds.
+ */
+static void measure_bogomips_ns(uint64_t iterations, int hart_count)
```
jserv, line 78: You should explain the rationale between BogoMIPS and the overhead of a HRT call.

utils.h
```diff
@@ -1,13 +1,16 @@
 #pragma once
 
+#include <stdbool.h>
 #include <stdint.h>
 
+extern bool boot_complete; /* Time to reach the first user process. */
```
jserv, line 6: Explain why we have to track the time to each the first user process.

jserv:
Running on macOS + Apple M1, I got the following:
```
$ make SMP=8 check
...
[    0.000000] rcu:     RCU restricting CPUs from NR_CPUS=32 to nr_cpu_ids=8.
[    0.000000] rcu: RCU calculated value of scheduler-enlistment delay is 25 jiffies.
[    0.000000] rcu: Adjusting geometry for rcu_fanout_leaf=16, nr_cpu_ids=8
...
[   31.988673] rcu: INFO: rcu_sched self-detected stall on CPU
[   31.993102] rcu:     2-....: (5247 ticks this GP) idle=2e0c/1/0x40000002 softirq=517/517 fqs=2128
[   32.002647]  (t=5253 jiffies g=-891 q=11 ncpus=8)
[   32.007524] CPU: 2 PID: 8 Comm: kworker/u16:0 Not tainted 6.1.99 #1
[   32.013259] Hardware name: semu (DT)
[   32.016691] Workqueue: events_unbound async_run_entry_fn
[   32.022636] epc : do_reset+0x30/0x6c
[   32.027939]  ra : do_reset+0x38/0x6c
[   32.033080] epc : c03816d0 ra : c03816d8 sp : c0893e50
[   32.038049]  gp : c04de840 tp : c085af40 t0 : c0bfa5d0
[   32.042925]  t1 : 00000008 t2 : 6d2e8da2 s0 : c0893e60
[   32.047778]  s1 : c03a48b8 a0 : 00000001 a1 : 00000000
[   32.052489]  a2 : 00000000 a3 : 005ac9b1 a4 : 0025364e
[   32.057287]  a5 : 00000000 a6 : 00000000 a7 : 0000000c
[   32.061854]  s2 : 004725a7 s3 : 0038da58 s4 : 00000000
[   32.066685]  s5 : 00000030 s6 : 00000007 s7 : c03a496c
[   32.071414]  s8 : c03a48b8 s9 : c0381488 s10: c03a4968
[   32.076499]  s11: c0381894 t3 : c0a7e400 t4 : 00000003
[   32.081382]  t5 : 46f17671 t6 : c0893e08
[   32.085321] status: 00000120 badaddr: dfcac9b1 cause: 80000005
[   32.090664] [<c03816d0>] do_reset+0x30/0x6c
[   32.096556] [<c0381870>] write_buffer+0x40/0x64
[   32.102722] [<c038207c>] unpack_to_rootfs+0x1c4/0x300
[   32.109299] [<c0382848>] do_populate_rootfs+0x70/0xd8
[   32.114948] [<c002bd8c>] async_run_entry_fn+0x30/0xd0
[   32.120732] [<c0021990>] process_one_work+0x198/0x224
[   32.126716] [<c0021f18>] worker_thread+0x238/0x294
[   32.132557] [<c00289c0>] kthread+0xd4/0xd8
[   32.137496] [<c0002514>] ret_from_exception+0x0/0x1c
[   55.544062] Freeing initrd memory: 8188K
```

jserv (changes requested):
Rebase the latest `master` branch and resolve the conflicts.

jserv (changes requested):
Rebase the latest `master` branch and resolve the conflicts.

Mes0903:
The current issue lies in the fact that the bogoMIPS measurement may have a significant discrepancy compared to the actual value. I suspect that this discrepancy arises because the instruction cache (i-cache) is in a warm state during the bogoMIPS measurement, whereas in real scenarios, this is not always the case. However, I have yet to come up with an effective method to measure and verify this hypothesis.

Regardless, if we continue to use the scale frequency method, the complexity of the program will undoubtedly increase. We would need to refine the measurement approach or implement additional corrections during the boot process to better align the values with real-world conditions.

Fundamentally, this complexity arises because the scale frequency method requires three key pieces of information to be implemented:
- The cost of `semu_timer_clocksource`
- The proportion of `semu_timer_clocksource`
- The number of times `semu_timer_clocksource` is called

Among these, the third piece of information is highly consistent across different environments, approximately equal to `SMP count * 2.15 * 1e8`. However, the other two values vary across different environments, which led to the introduction of bogoMIPS in the first place, as we needed a way to measure these two variables. Unfortunately, our current measurement approach does not accurately reflect real-world conditions.

For example, in an M2 Pro Max environment with SMP set to 32, the relationship between the values is illustrated in the figure below:

image

The red dashed line represents the value measured by bogoMIPS, while the blue line represents the actual value. The significant gap between the two results in the persistent occurrence of RCU CPU stall warnings.

I have re-evaluated the increment method. Since it only requires the number of `semu_timer_clocksource` calls for implementation and this value remains highly consistent across different environments:

image

As expected, this method effectively suppresses RCU CPU stall warnings across various environments.

Initially, we chose the scale frequency method because:
- Using a hardcoded `scale_factor` made implementation relatively simple.
- It provided an intuitive way to estimate the actual boot time (e.g., if `scale_factor` was set to 0.01, the actual boot time would be the printed value multiplied by 100).

However, after a series of modifications, these advantages no longer hold. Additionally, the implementation of the scale frequency method has become more complex than the increment method. Furthermore, the increment method eliminates the need to call `clock_gettime`, providing another advantage.

Therefore, I think we should switch to the increment method to resolve the RCU CPU stall warning issue.

PR #70, opened 2025-01-26
Implement VirtIO RNG device

### LINKED ISSUE #68: VirtIO RNG and unexpected behavior of arc4random_buf() standard function of C
This issue was identified during the development of #34 , where [kmscube](https://gitlab.freedesktop.org/mesa/kmscube) was found to have an extremely high chance of hanging during startup.

Upon investigation, I discovered that the reason kmscube hangs is due to the use of [arc4random_buf()](https://man.freebsd.org/cgi/man.cgi?query=arc4random_buf&sektion=3&n=1), a standard C library function. Since Buildroot packages are mostly built from source, I was able to trace the issue directly to its source.

Firstly, kmscube has the following package dependencies:
`kmscube` -> `Mesa3D` -> `libgbm` (Generic Buffer Management) -> [libexpat](https://github.com/libexpat/libexpat) (an XML parser)

During its startup, kmscube calls a libgbm function that reads application and driver settings from [00-mesa-defaults.conf](https://github.com/intel/external-mesa/blob/master/src/util/00-mesa-defaults.conf). This process involves libexpat to parse XML syntax.

It turns out that during initialization, the parser calculates a salt hash, which requires using `arc4random_buf()` to generate a random number. However, due to an unknown issue, `arc4random_buf()` often hangs, preventing kmscube from starting.

Currently, we can work around the issue by modifying the following file:

> **buildroot/output/build/expat-2.6.2/lib/xmlparse.c**

```diff
static unsigned long
generate_hash_secret_salt(XML_Parser parser) {
+ return 2147483647;

  unsigned long entropy;
  (void)parser;

  /* "Failproof" high quality providers: */
#if defined(HAVE_ARC4RANDOM_BUF)
  arc4random_buf(&entropy, sizeof(entropy));
  return ENTROPY_DEBUG("arc4random_buf", entropy);
#elif defined(HAVE_ARC4RANDOM)
  ...
}
```

shengwen-tw:
## Overview

This commit introduces the VirtIO entropy device (also know as virtio-rng in QEMU and the Linux kernel) to resolve the blocking issue of `arc4random_buf()` [[1]](https://elixir.bootlin.com/glibc/glibc-2.36/source/stdlib/arc4random.c) caused by insufficient entropy of `/dev/random`.

According to the man page (`man 7 random`):

> The kernel random-number generator relies on entropy gathered from device drivers and other sources of environmental noise to seed a cryptographically secure pseudorandom number generator (CSPRNG).
>
>**Interface Pool:** /dev/random
>**Pool:** Blocking pool
>**Blocking behavior:** If entropy too low, blocks until there is enough entropy
>**Behavior when pool is not yet ready:** Blocks until enough entropy gathered

Quaoted from https://en.wikipedia.org/wiki//dev/random

> With Linux kernel 3.16 and newer, the kernel itself mixes data from hardware random number generators into /dev/random on a sliding scale based on the definable entropy estimation quality of the HWRNG. This means that no userspace  daemon, such as rngd from rng-tools, is needed to do that job. With Linux kernel 3.17+, the VirtIO RNG was modified to have a default quality defined above 0, and as such, is currently the only HWRNG mixed into /dev/random by default.

[1] https://elixir.bootlin.com/glibc/glibc-2.36/source/stdlib/arc4random.c

Close #68.

## Prerequisites

1. Enable `CONFIG_OD` in BusyBox

```shell
$ cd buildroot/output/build/busybox-1.36.1/
$ make menuconfig
```

2. Rebuild buildroot

```shell
$ cd buildroot/
$ make busybox-rebuild
$ make -j$(nproc)
```

3. Replace `rootfs.cpio`

```shell
$ cd semu
$ cp -f buildroot/output/images/rootfs.cpio ./
```

## Test procedures

1. Launch `semu`:

```shell
cd semu/
make check
```

2. Check if `virtio-rng` is used

```
# cat /sys/class/misc/hw_random/rng_available
virtio_rng.0
```

```
# cat /sys/class/misc/hw_random/rng_current
virtio_rng.0
```

3. Read `/dev/random` with `od` command

```
Welcome to Buildroot
buildroot login: root
# od /dev/random
0000000 126370 025274 055514 014745 051051 073714 156246 140301
0000020 157561 167176 122602 015767 107243 120045 136313 176061
0000040 115122 170566 107766 011777 057377 105023 152157 147260
0000060 130220 064322 160762 057245 153463 141757 046574 122202
0000100 174113 136630 120230 143307 073562 077302 017507 073016
0000120 062537 012454 100047 114555 140204 040440 172507 075066
0000140 040442 073115 161077 100450 037510 166327 163375 133510
0000160 176100 103667 010651 077154 077252 122605 046415 067073
0000200 042206 016216 101214 106563 161454 053741 016273 031322
0000220 166247 017356 066413 041563 124027 031751 054160 122432
...
```

Without `virtio-rng`, the `od` command will just hang due to the aforementioned issue.


ChinYikMing:
Can you upload the newly built rootfs.cpio here for quick testing? Thanks!

shengwen-tw:
> Can you upload the newly built rootfs.cpio here for quick testing? Thanks!

You could try this: [rootfs.cpio.zip](https://github.com/user-attachments/files/18554122/rootfs.cpio.zip)

jserv (changes requested):
Improve the descriptions in git commit messages by quoting the man page for entropy. Also, always use `Close #68` instead of `Closed #68` to emphasize the intention.

ChinYikMing:
I try on machine x86-64, compiled with gcc 12.3.0.

Procedure:
1. `$ make check`
2. login with root
3. `# od /dev/random`

Result:
```
...

[    3.884043] Run /init as init process
Starting syslogd: OK
Starting klogd: OK
Running sysctl: OK
Starting network: OK

Welcome to Buildroot
buildroot login: root 
# od /dev/random



```

Nothing is output (blocked), is it normal? I think the `/dev/random` should not be blocked?

shengwen-tw:
> I try on machine x86-64, compiled with gcc 12.3.0.
> 
> Procedure:
> 
> 1. `$ make check`
> 2. login with root
> 3. `# od /dev/random`
> 
> Result:
> 
> ```
> ...
> 
> [    3.884043] Run /init as init process
> Starting syslogd: OK
> Starting klogd: OK
> Running sysctl: OK
> Starting network: OK
> 
> Welcome to Buildroot
> buildroot login: root 
> # od /dev/random
> ```
> 
> Nothing is output (blocked), is it normal? I think the `/dev/random` should not be blocked?

It should not be blocked, could you check the following commands?

```shell
cat /sys/class/misc/hw_random/rng_available
virtio_rng.0
```

```shell
cat /sys/class/misc/hw_random/rng_current
virtio_rng.0
```

virtio-rng.c
```diff
@@ -0,0 +1,268 @@
...
+        return;
+    default:
+        vm_set_exception(vm, RV_EXC_ILLEGAL_INSN, 0);
+        return;
+    }
+}
+
+void virtio_rng_init(void)
+{
+    rng_fd = open("/dev/urandom", O_RDONLY);
```
jserv, line 263: `/dev/random` is probably going to be sufficient. What are the considerations to use `/dev/urandom`?

shengwen-tw: I used `/dev/urandom` because it does not block. However, this concern may not arise on the host, and adopting `/dev/random` is indeed more secure.

ChinYikMing:
> > I try on machine x86-64, compiled with gcc 12.3.0.
> > Procedure:
> > 
> > 1. `$ make check`
> > 2. login with root
> > 3. `# od /dev/random`
> > 
> > Result:
> > ```
> > ...
> > 
> > [    3.884043] Run /init as init process
> > Starting syslogd: OK
> > Starting klogd: OK
> > Running sysctl: OK
> > Starting network: OK
> > 
> > Welcome to Buildroot
> > buildroot login: root 
> > # od /dev/random
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Nothing is output (blocked), is it normal? I think the `/dev/random` should not be blocked?
> 
> It should not be blocked, could you check the following commands?
> 
> ```
> # cat /sys/class/misc/hw_random/rng_available
> virtio_rng.0
> ```
> 
> ```
> # cat /sys/class/misc/hw_random/rng_current
> virtio_rng.0
> ```

I get the following output:
```
# cat /sys/class/misc/hw_random/rng_available
cat: can't open '/sys/class/misc/hw_random/rng_available': No such file or directory
```

I noticed that the Linux kernel needs to enable the virtio-rng driver to facilitate communication in this scenario (enabled in the proposed changes of Linux configuration). Could you also upload the virtio-rng enabled Linux kernel image along with the rootfs.cpio?

shengwen-tw:
> I noticed that the Linux kernel needs to enable the virtio-rng driver to facilitate communication in this scenario (enabled in the proposed changes of Linux configuration). Could you also upload the virtio-rng enabled Linux kernel image along with the rootfs.cpio?

Here you go: [Image.zip](https://github.com/user-attachments/files/18589587/Image.zip)

ChinYikMing:
Using newly built Linux kernel, the `od /dev/random` has no more blocking.

ChinYikMing:
BTW, I used the newly built rootfs.cpio within [rv32emu](https://github.com/sysprog21/rv32emu) which does not implement the VirtIO RNG device and it does not block the `/dev/random` by running `od /dev/random`. @shengwen-tw Do you have any idea with this?

virtio-rng.c
```diff
@@ -0,0 +1,268 @@
...
+    struct virtq_desc vq_desc = {
+        .addr = desc[0],
+        .len = desc[2],
+        .flags = desc[3],
+    };
+
+    /* Write entropy buffer */
+    void *entropy_buf =
+        (void *) ((uintptr_t) vrng->ram + (uintptr_t) vq_desc.addr);
+    ssize_t total = read(rng_fd, entropy_buf, vq_desc.len);
```
ChinYikMing, line 74: `read` could fail. Please handle the return value.

shengwen-tw:
> BTW, I used the newly built rootfs.cpio within [rv32emu](https://github.com/sysprog21/rv32emu) which does not implement the VirtIO RNG device and it does not block the `/dev/random` by running `od /dev/random`. @shengwen-tw Do you have any idea with this?

Entropy collection can be influenced by the speed of the hardware or the emulated environment.
I believe that rv32emu is currently fast enough to prevent blocking. However, it is always nice to have entropy supplied by the host.

virtio-rng.c
shengwen-tw: According to `man 4 random`

>When the entropy pool is empty, reads from /dev/random will block until additional environmental noise  is  gathered.

Calling `read()` function on `/dev/random` can only return `-1` (i.e., failed) if `O_NONBLOCK` flag is set when opening the device. Specifically:

>If  open(2)  is called for /dev/random with the O_NONBLOCK flag, a subsequent read(2) will not block if the requested number of bytes is not available. Instead, the available bytes are returned.  If no byte is available, read(2) will return -1 and errno will be set to EAGAIN.

PR #72, opened 2025-02-01
Rectify length of Virtio descriptor address

otteryc:
According to Virtio Specification Version 1.3, Section 2.7.5 *The Virtqueue Descriptor Table*, the type of `addr` should be `uint64_t` instead of `uint32_t`.
This commit addresses the error in the previous implementation.

### DIFF
--- a/virtio-rng.c
+++ b/virtio-rng.c
@@ -61,20 +61,16 @@ static void virtio_queue_notify_handler(virtio_rng_state_t *vrng,
     VRNG_QUEUE.last_avail++;
 
     /* Read descriptor */
-    uint32_t *desc = &vrng->ram[queue->QueueDesc + buffer_idx * 4];
-    struct virtq_desc vq_desc = {
-        .addr = desc[0],
-        .len = desc[2],
-        .flags = desc[3],
-    };
+    struct virtq_desc *vq_desc =
+        (struct virtq_desc *) &vrng->ram[queue->QueueDesc + buffer_idx * 4];
 
     /* Write entropy buffer */
     void *entropy_buf =
-        (void *) ((uintptr_t) vrng->ram + (uintptr_t) vq_desc.addr);
-    ssize_t total = read(rng_fd, entropy_buf, vq_desc.len);
+        (void *) ((uintptr_t) vrng->ram + (uintptr_t) vq_desc->addr);
+    ssize_t total = read(rng_fd, entropy_buf, vq_desc->len);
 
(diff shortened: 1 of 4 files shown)


virtio-rng.c
```diff
@@ -61,20 +61,16 @@ static void virtio_queue_notify_handler(virtio_rng_state_t *vrng,
...
 
     /* Read descriptor */
-    uint32_t *desc = &vrng->ram[queue->QueueDesc + buffer_idx * 4];
-    struct virtq_desc vq_desc = {
-        .addr = desc[0],
-        .len = desc[2],
-        .flags = desc[3],
-    };
+    struct virtq_desc vq_desc =
+        *(struct virtq_desc *) &vrng->ram[queue->QueueDesc + buffer_idx * 4];
```
shengwen-tw, line 65: Use `struct *virtq_desc vq_desc` instead of `struct virtq_desc vq_desc` to ensure the descriptor is updated for the driver (Guest OS) rather than just being copied locally.

virtio-rng.c
```diff
@@ -61,20 +61,16 @@ static void virtio_queue_notify_handler(virtio_rng_state_t *vrng,
...
 
     /* Write entropy buffer */
     void *entropy_buf =
         (void *) ((uintptr_t) vrng->ram + (uintptr_t) vq_desc.addr);
     ssize_t total = read(rng_fd, entropy_buf, vq_desc.len);
 
     /* Clear write flag */
-    desc[3] = 0;
+    vq_desc.flags = 0;
```
shengwen-tw, line 73: Should be `vq_desc->flags = 0;`

virtio.h
```diff
@@ -58,7 +58,7 @@ enum {
 };
 
 struct virtq_desc {
```
shengwen-tw, line 60: Let's wrap the `struct virtq_desc` with [PACKED](https://github.com/sysprog21/semu/blob/d61c9f6e337e52407a7ac22739d3f145e397fc02/common.h#L31) macro explicitly.

shengwen-tw (changes requested):

jserv (changes requested):
Avoid unnecessary capitalization. That is, use "Rectify length of Virtio descriptor address" instead of "Rectify Length of Virtio Descriptor Address."

shengwen-tw (approved):
Looks legitimate.

jserv:
Thank @otteryc for contributing!

PR #74, opened 2025-02-06
Support GDB stub for remote debugging

RinHizakura:
Enable the emulator to act as gdbstub. This feature helps to dignose and debug issue of emulator more simply. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements GDB remote debugging functionality by integrating mini-gdbstub library, adding debug mode with -g option, and restructuring the main emulation loop. The changes enable the emulator to function as a GDB remote target on port 1234, supporting interactive debugging through register/memory access and execution control features.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 4
</div>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,7 @@
     path = cnfa
     url = https://github.com/cntools/cnfa
     shallow = true
+[submodule "mini-gdbstub"]
+    path = mini-gdbstub
+    url = https://github.com/RinHizakura/mini-gdbstub
+    shallow = true
--- a/main.c
+++ b/main.c
@@ -1,4 +1,5 @@
 #include <assert.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
 #include <stdio.h>
@@ -9,9 +10,9 @@
 #include <unistd.h>
 
 #include "device.h"
+#include "mini-gdbstub/include/gdbstub.h"
 #include "riscv.h"
 #include "riscv_private.h"
-
 #define PRIV(x) ((emu_state_t *) x->priv)
 
 /* Define fetch separately since it is simpler (fixed width, already checked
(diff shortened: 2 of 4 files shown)


main.c
```diff
@@ -591,158 +596,303 @@ static void handle_options(int argc,
...
+    emu_state_t *emu = &semu->emu;
+    vm_t *vm = &semu->vm;
     handle_options(argc, argv, &kernel_file, &dtb_file, &initrd_file,
-                   &disk_file, &netdev, &hart_count);
+                   &disk_file, &netdev, &hart_count, &debug);
 
     /* Initialize the emulator */
-    emu_state_t emu;
-    memset(&emu, 0, sizeof(emu));
+    memset(emu, 0, sizeof(*emu));
```
jserv, line 614: Check the comments in `riscv.h` carefully:
```c
/* To use the emulator, start by initializing a hart_t object with zero values,
 * invoke vm_init(), and set the required environment-supplied callbacks. You
 * may also set other necessary fields such as argument registers and s_mode,
 * ensuring that all field restrictions are met to avoid undefined behavior.
 *
 * Once the emulator is set up, execute the emulation loop by calling
 * "vm_step()" repeatedly. Each call attempts to execute a single instruction.
 *
 * If the execution completes successfully, the "vm->error" field will be set
 * to ERR_NONE. However, if an error occurs during execution, the emulator will
 * halt and the "vm->error" field will provide information about the error. It
 * is important to handle the emulation error before calling "vm_step()" again;
 * otherwise, it will not execute any instructions. The possible errors are
 * described above for reference.
 */
```

Do follow the manner unless you are about to refine the usage as the major purpose.

RinHizakura: The purpose of abstracting `semu_init()` here is to share the same logic between debug & normal mode.  The behavior in normal mode should not be changed here.

RinHizakura:
I think the change still respects the riscv.h manner, as no functional change for the normal mode. The commit just separates the initialization of semu to reuse it for debug mode. Please let me know if you still need any change from me .

main.c
```diff
@@ -591,158 +596,303 @@ static void handle_options(int argc,
...
+static int semu_init(semu_t *semu, int argc, char **argv)
 {
     char *kernel_file;
     char *dtb_file;
     char *initrd_file;
     char *disk_file;
     char *netdev;
     int hart_count = 1;
+    bool debug = false;
+    emu_state_t *emu = &semu->emu;
```
jserv, line 608: The statement `emu_state_t *emu = &semu->emu;` does not look elegant since adjacent `semu` and `emu` cause confusion. Can you refine it?

.gitmodules
```diff
@@ -2,3 +2,7 @@
     path = cnfa
     url = https://github.com/cntools/cnfa
     shallow = true
+[submodule "mini-gdbstub"]
+    path = mini-gdbstub
+    url = https://github.com/RinHizakura/mini-gdbstub.git
```
jserv, line 7: Drop the suffixing `.git`.

main.c
RinHizakura: The variable name `semu` for `semu_t` is very intuitive, and I don't want to change `emu` to other names as it used to define like. The solution I can come up with here is to change `semu_t` to other names. Any suggestion?

jserv: How about consolidating the existing `semu` structure instead?

RinHizakura:
Integrate all fields include `vm_t` into `emu_state_t`, means to use `emu_state_t` to represent the main body of semu

jserv:
Thank @RinHizakura for contributing!

PR #75, opened 2025-02-14
Manage ticks to suppress RCU CPU stall warning

### LINKED ISSUE #51: RCU CPU stall warning in a multi-core system simulation
Execute semu with multi-core system simulation
```
make check SMP=4
```

The [RCU CPU stall warning], as discussed in #49 , is accompanied by an increase in timer interrupts and `clock_gettime` system calls to produce a real-time timer. This causes the CPU to wait longer than a typical grace period, which is usually 21 seconds.

```
[   33.044071] rcu: INFO: rcu_sched self-detected stall on CPU
[   33.048071] rcu: 	2-....: (5247 ticks this GP) idle=da1c/1/0x40000002 softirq=689/689 fqs=1847
[   33.056071] 	(t=5253 jiffies g=-887 q=11 ncpus=4)
[   33.064071] CPU: 2 PID: 8 Comm: kworker/u8:0 Not tainted 6.1.94 #12
[   33.068071] Hardware name: semu (DT)
[   33.072071] Workqueue: events_unbound async_run_entry_fn
[   33.080071] epc : eat+0x34/0x54
[   33.088071]  ra : do_reset+0x38/0x6c
[   33.096071] epc : c03811fc ra : c038137c sp : c0871e40
[   33.104071]  gp : c04de858 tp : c0852f40 t0 : c0aabc70
[   33.112071]  t1 : 00000008 t2 : 67e0cb47 s0 : c0871e50
[   33.120071]  s1 : c03a48b8 a0 : 00000001 a1 : 00000000
[   33.128071]  a2 : 00000000 a3 : 0053be47 a4 : 00000000
[   33.136071]  a5 : c03a48b8 a6 : 00000000 a7 : 0000000c
[   33.144071]  s2 : 004725a7 s3 : 0038da58 s4 : 00000000
[   33.152071]  s5 : 00000030 s6 : 00000007 s7 : c03a496c
[   33.156071]  s8 : c03a48b8 s9 : c038112c s10: c03a4968
[   33.160071]  s11: c0381538 t3 : c09e4140 t4 : 00000003
[   33.168071]  t5 : ec37d331 t6 : c0871e08
[   33.172071] status: 00000120 badaddr: c03a48c0 cause: 80000005
[   33.180071] [<c03811fc>] eat+0x34/0x54
[   33.184071] [<c038137c>] do_reset+0x38/0x6c
[   33.192071] [<c0381514>] write_buffer+0x40/0x64
[   33.200071] [<c0381d20>] unpack_to_rootfs+0x1c4/0x300
[   33.212072] [<c03824ec>] do_populate_rootfs+0x70/0xd8
[   33.220072] [<c002bd58>] async_run_entry_fn+0x30/0xd0
[   33.232072] [<c002195c>] process_one_work+0x198/0x224
[   33.240072] [<c0021ee4>] worker_thread+0x238/0x294
[   33.248072] [<c002898c>] kthread+0xd4/0xd8
[   33.252072] [<c0002508>] ret_from_exception+0x0/0x1c
```

Implementing multi-thread support for semu, as discussed in PR #49, might improve the performance of the booting process.

[RCU CPU stall warning]: https://docs.kernel.org/RCU/stallwarn.html

Mes0903:
Since the emulator currently operates using sequential emulation, the execution time for the boot process is relatively long, which can result in the generation of RCU CPU stall warnings.

To address this issue, there are several potential solutions:

1. Scale the frequency to slow down emulator time during the boot process, thereby eliminating RCU CPU stall warnings.
2. During the boot process, avoid using 'clock_gettime' to update ticks and instead manage the tick increment relationship manually.
3. Implement multi-threaded emulation to accelerate the emulator's execution speed.

For the third point, while implementing multi-threaded emulation can significantly accelerate the emulator's execution speed, it cannot guarantee that this issue will not reappear as the number of cores increases in the future. Therefore, a better approach is to use methods 1 and 2 to allow the emulator to set an expected time for completing the boot process.

The scale method requires three pieces of information to be implemented:
- the cost of 'semu_timer_clocksource'
- percentage of 'semu_timer_clocksource' within the boot process
- the number of times 'semu_timer_clocksource' is called.

In contrast, the increment method only requires the third information to be implemented, making its implementation simpler.

Furthermore, through statistical analysis, we found that the other two values (cost and percentage) exhibit different distributions across different environments. Therefore, using the scale method would require an additional profiling step when the emulator starts, adding complexity to the implementation.

Finally, since the increment method does not obtain real-time timestamps during the boot process, its overhead is lower compared to the scale method. Because 'semu_timer_clocksource' is called frequently, this reduction in overhead accumulates and results in a noticeable performance improvement.

Therefore, this commit opts for the increment method to address this issue. This commit divides time into emulator time and real time. During the boot process, the timer manage the ticks increment to slow down the growth of emulator time, eliminating RCU CPU stall warnings. After the boot process is complete, the growth of emulator time aligns with real time.

According to Using RCU’s CPU Stall Detector [1], the grace period for RCU CPU stalls is typically set to 21 seconds. By dividing this value by two as the expected completion time, we can provide a sufficient buffer to reduce the impact of errors and avoid RCU CPU stall warnings.

By statistic, the number of times 'semu_timer_clocksource' called is approximately

$$
\text{SMP count} \times 2.15 \times 10^8
$$

By the time the boot process is completed, the emulator will have a total of 

$$
\text{boot seconds} \times \text{frequency}
$$

ticks. Therefore, each time, 

$$
\frac{\text{boot seconds} \times \text{frequency}}{(2.15 \times 10^8 \times \text{SMP count})}
$$

ticks need to be added.

Close #51

[1] https://docs.kernel.org/RCU/stallwarn.html#config-rcu-cpu-stall-timeout 
 <div id='description'>
<h3>Summary by Bito</h3>
Implementation of a sophisticated timer management system with dual-phase timing to prevent RCU CPU stall warnings during boot process. The system features manual tick increment management during boot, transitioning to real-time timing after detecting boot completion via U mode transition. The solution includes statistical calculations for optimal tick increments based on hart count, creating a robust timing infrastructure.
<br>
<br>
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 2
</div>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,16 @@ DTC ?= dtc
 E :=
 S := $E $E
 
+# During boot process, he emulator manually manages the growth of ticks to 
+# suppress RCU CPU stall warnings. Thus, we need an target time to set the 
+# increment of ticks. According to Using RCU’s CPU Stall Detector[1], the
+# grace period for RCU CPU stalls is typically set to 21 seconds. 
+# By dividing this value by two as the expected completion time, we can 
+# provide a sufficient buffer to reduce the impact of errors and avoid 
+# RCU CPU stall warnings.
+# [1] docs.kernel.org/RCU/stallwarn.html#config-rcu-cpu-stall-timeout
+CFLAGS += -D SEMU_BOOT_TARGET_TIME=10
+
 SMP ?= 1
 .PHONY: riscv-harts.dtsi
 riscv-harts.dtsi:
--- a/main.c
+++ b/main.c
@@ -678,7 +678,7 @@ static int semu_init(emu_state_t *emu, int argc, char **argv)
     virtio_rng_init();
 #endif
     /* Set up ACLINT */
-    semu_timer_init(&emu->mtimer.mtime, CLOCK_FREQ);
+    semu_timer_init(&emu->mtimer.mtime, CLOCK_FREQ, hart_count);
     emu->mtimer.mtimecmp = calloc(vm->n_hart, sizeof(uint64_t));
     emu->mswi.msip = calloc(vm->n_hart, sizeof(uint32_t));
     emu->sswi.ssip = calloc(vm->n_hart, sizeof(uint32_t));
--- a/riscv.c
+++ b/riscv.c
@@ -382,6 +382,14 @@ static void op_sret(hart_t *vm)
     vm->s_mode = vm->sstatus_spp;
     vm->sstatus_sie = vm->sstatus_spie;
 
+    /* After the booting process is complete, initrd will be loaded. At this
+     * point, the sytstem will switch to U mode for the first time. Therefore,
+     * by checking whether the switch to U mode has already occurred, we can
+     * determine if the boot process has been completed.
+     */
+    if (!boot_complete && !vm->s_mode)
+        boot_complete = true;
+
     /* Reset stack */
     vm->sstatus_spp = false;
     vm->sstatus_spie = true;
--- a/utils.c
+++ b/utils.c
@@ -1,3 +1,4 @@
+#include <stdbool.h>
 #include <time.h>
 
 #include "utils.h"
@@ -19,6 +20,10 @@
 #endif
 #endif
 
+bool boot_complete = false;
+static double ticks_increment;
+static double boot_ticks;
+
 /* Calculate "x * n / d" without unnecessary overflow or loss of precision.
  *
  * Reference:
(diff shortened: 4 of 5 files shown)


jserv:
`clang-format` version 12 would incorrectly indent as following:
```diff
--- utils.h
+++ utils.h
@@ -87,7 +87,7 @@ static inline void list_del_init(struct
     })
 #else
 #define container_of(ptr, type, member) \
-    ((type *) ((char *) (ptr) - offsetof(type, member)))
+    ((type *) ((char *) (ptr) -offsetof(type, member)))
 #endif
 
 #define list_entry(node, type, member) container_of(node, type, member)
```

We shall migrate to newer versions later.

Makefile
```diff
@@ -133,6 +133,7 @@ E :=
 S := $E $E
 
 SMP ?= 1
+CFLAGS += -D SEMU_BOOT_TARGET_TIME=10
```
jserv, line 136: Add proper comments.

jserv:

PR #79, opened 2025-04-16
Update mini-gdbstub with the API changes

RinHizakura:
To support register >64bits(e.g. AVX register on x86), the corresponding interface changes are made on mini-gdbstub. Synchronize the library and interface to the latest version to facilitate future development. This may also be helpful if we had Vector Extensions in the future.

Similar change made for rv32emu: https://github.com/sysprog21/rv32emu/pull/589


jserv:
Thank @RinHizakura for contributing!

PR #97, opened 2025-10-23
Add coroutine-based SMP support with WFI

jserv:
This implements cooperative multitasking for multi-hart systems using coroutines, enabling efficient SMP emulation with significant CPU usage reduction.
- WFI instruction callback mechanism for power management
- CPU usage optimization: ~90% reduction in idle systems
- Maximum latency: 1ms (acceptable for typical 10ms timer interrupts)
    

---

## Summary by cubic
Adds coroutine-based cooperative SMP for multi-hart emulation. Harts yield on WFI to cut idle CPU usage by ~90%, with up to 1ms latency.

- **New Features**
  - Lightweight coroutine runtime (x86_64/ARM64 assembly, ucontext fallback).
  - Per-hart coroutines and round-robin scheduling in semu_run.
  - WFI callback to suspend a hart; sleep 1ms when all started harts are waiting.
  - Hart execution loop with batched steps and proper trap/ECALL handling.
  - Makefile adds coro.o; hart0 PC explicitly set to 0x0.

- **Bug Fixes**
  - Initialize ppn in mmu_translate to avoid undefined behavior.

### DIFF
--- a/coro.c
+++ b/coro.c
@@ -234,6 +234,13 @@ static __thread coro_t *tls_running_coro = NULL;
 static coro_t *tls_running_coro = NULL;
 #endif
 
+static inline void coro_clear_running_state(void)
+{
+    coro_state.current_hart = CORO_HART_ID_IDLE;
+    coro_state.running = NULL;
+    tls_running_coro = NULL;
+}
+
 /* Forward declarations */
 
 #ifdef CORO_USE_ASM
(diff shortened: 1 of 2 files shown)


shengwen-tw:
> > This looks interesting, but wouldn’t using a real thread be faster than a coroutine for HART emulation? Is this approach chosen to keep the simplicity of semu?
> 
> Before refining the internal structure toward multiple harts, let’s stick to single-threaded emulation as early versions of QEMU did.

I see, I think this is an interesting one to follow.

PR #111, opened 2025-11-02
Fix SMP=4 boot hang regression

jserv:
## Problem

#110 introduced a regression where SMP=4 configurations hang during boot after SMP initialization completes.

### Symptoms
- **SMP=4**: Hung at `[0.049151] smp: Brought up 1 node, 4 CPUs` (49 lines of output)
  - Never reached `clocksource: Switched to clocksource` or login prompt
- **SMP=1**: Continued to work correctly
- **Regression introduced**: PR #110 (commit 1aab2cb)
- **Last working commit**: 4552c62 (pre-coro-uart)

## Root Cause

The adaptive timer/UART fd registration logic excluded timer fd from poll() monitoring when all harts entered WFI state. During early boot, the kernel temporarily puts all harts in WFI while waiting for timer interrupts. This created a deadlock:

1. All 4 harts enter WFI waiting for timer interrupts
2. Logic sees `idle_harts == started_harts`, sets `harts_active = false`  
3. Timer fd not added to poll() fds
4. poll() blocks indefinitely (or sleeps 10ms repeatedly)
5. Timer interrupts never fire → harts never wake → deadlock

## Fix

Introduce boot completion heuristic using `peripheral_update_ctr`:

```c
const uint64_t BOOT_SETTLE_ITERATIONS = 5000;
bool boot_complete = all_harts_started && 
                     (emu->peripheral_update_ctr > BOOT_SETTLE_ITERATIONS);
bool harts_active = (vm->n_hart == 1) || !boot_complete || (idle_harts == 0);
```

During the first 5000 scheduler iterations after SMP initialization, always keep timer and UART fds active. This ensures harts can receive timer interrupts even when temporarily in WFI during early boot.

---
## Summary by cubic
Fixes the SMP=4 boot hang introduced in PR #110 by keeping timer and UART polling active during early boot to avoid a WFI deadlock. Boots reliably on SMP=1 and SMP=4, with no impact to post-boot idle optimizations.

- **Bug Fixes**
  - Root cause: timer fd was dropped from poll() when all harts entered WFI during early boot, blocking timer interrupts.
  - Change: add a boot completion heuristic using peripheral_update_ctr with a 5000-iteration settle period; treat boot as incomplete and keep timer/UART fds active until then or while any hart is active.
  - Result: SMP=4 now completes boot; SMP=1 remains stable.
  - Performance: preserves PR #110’s idle CPU optimization after boot.

<sup>Written for commit 523748ede1c41275fa771cfcdf0d8698ab714abf. Summary will update automatically on new commits.</sup>

### DIFF
--- a/main.c
+++ b/main.c
@@ -1191,22 +1191,20 @@ static int semu_run(emu_state_t *emu)
             size_t pfd_count = 0;
             int timer_index = -1;
 
-            /* Add periodic timer fd (1ms interval for guest timer emulation).
-             * Only add timer when ALL harts are active (none idle) to allow
-             * poll() to sleep when any harts are in WFI. When harts are idle,
-             * timer updates can be deferred until they wake up.
-             *
-             * During SMP boot (started_harts < vm->n_hart), always include the
-             * timer to ensure secondary harts can complete initialization. Only
-             * apply conditional exclusion after all harts have started.
-             *
-             * For single-hart configurations (n_hart == 1), disable
-             * optimization entirely to avoid boot issues, as the first hart
-             * starts immediately.
+            /* Add periodic timer fd (1ms interval). Excluded when harts are
+             * idle to allow poll() sleep, but always included during:
+             * 1. Single-hart mode (n_hart == 1)
+             * 2. Boot phase (!boot_complete) - prevents deadlock when kernel
+             *    briefly puts all harts in WFI while waiting for timer IRQ
+             * 3. Active execution (idle_harts == 0)
              */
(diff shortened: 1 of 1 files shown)


jserv (commented):
Thank you for catching this! You're absolutely correct - `peripheral_update_ctr` is a countdown timer (64→0→64) and will never exceed 5000, meaning `boot_complete` stays false permanently. This breaks the intended idle optimization.

I will fix this by introducing a monotonic `scheduler_iterations` counter to properly track boot completion.

PR #34, opened 2023-11-09
Implement VirtIO GPU and input devices

shengwen-tw:
## Summary

This pull request introduces initial support for `virtio-gpu` and `virtio-input` devices in `semu`, conforming to the [VirtIO specification 1.3](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html).

### VirtIO-GPU

* Based on `VirtIO MMIO transport layer` (i.e., Linux platform device)
* Implements basic 2D rendering support.
* Provides two rendering backends:
    * **Software-based** backend (supports 2D rendering only, ready to be integrated into [rv32emu](https://github.com/sysprog21/rv32emu)).
    * **virglrenderer** backend (to enable hardware-accelerated 3D rendering in a future patch).
* Compatible with:
    * [DirectFB2](https://github.com/directfb2/DirectFB2)
    * [X11 Window System](https://www.x.org/wiki/)
* Successfully tested with:
    * [kmscube](https://gitlab.freedesktop.org/mesa/kmscube)
    * [glmark2](https://github.com/glmark2/glmark2)
    * [glxgears](https://linuxreviews.org/Glxgears#google_vignette)

### VirtIO-Input

* Implements basic keyboard emulation.
* Implements basic mouse emulation.

## Prerequisites

> Note: this list is not yet exhaustive

```shell
$ sudo apt install libsdl2-dev libsdl2-2.0-0 libsdl2-image-dev libsdl2-image-2.0-0
```

## Build kernel image and test rootfs image

```shell
$ make rootfs.cpio # Download prebuilt minimal rootfs for initrd
$ ./scripts/build-image.sh --all --x11 --external-root --extra-packages
```

Use `--x11` for full setup; omit it for a minimal setup.

## Run `X11`

1. Launch `semu`:

```shell
$ make check
```

2. Run `startx`

```
$ startx
```

3. After `X11` shows up, launch the benchmark programs:

```
$ glmark2
$ glxgears
```

## Run `DirectFB2`

1. Launch `semu`:

```shell
$ make check
```

2. Kill `X11` (Skip if `--x11` is not used)

```
$ ps | grep 'Xorg'
   <PID> root     /usr/bin/Xorg :0.0 vt01 -s 0 -noreset -allowMouseOpenFail
$ kill <PID>
```

3. Set up environment variables via `run.sh`:

```shell
$ source ./run.sh
```

4. Run `DirectFB2` test programs:

```
df_andi        df_fire df_neo     df_window
[...]

Issue #32, opened 2023-10-01
Implement VirtIO GPU device


shengwen-tw:
Currently semu does not support graphics rendering.

I think VirtIO GPU with 2D mode might be a good starting point to go. 
Later we can then emulate mouse and keyboard more sophisticated features.

jserv:
[homebrew-qemu-virgl](https://github.com/knazarov/homebrew-qemu-virgl) illustrates how virtio-gpu can be implemented via [VirGL](https://docs.mesa3d.org/drivers/virgl.html).

shengwen-tw:
**Side note:**
For implementing VirtIO GPU with **Linux 6.x kernel**, we should refer to [VIRTIO Spec. 1.2](https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html) (Draft 01 released on 09 May 2022) instead of earlier versions (i.e., [VirtIO Spec. v1.1](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-3200007)).

The register **0xb0**, **0xb4**, **0xb8**, **0xbc** and **0xc0** are only documented since v1.2.

shengwen-tw:
By observation, when the Linux kernel loads **virtio-gpu**  module, it reads register **0xac (SHMSel)**, **0xb0 (SHMLenLow)**, **0xb4 (SHMLenHigh)**, **0xb8 (SHMBaseLow)**, and **0xbc (SHMBaseHigh)** for a shared memory space. (Those registers tell the address and length of a shared memory specified with an ID.)

Seems like shared memory is meant to be a **device** by design, and may be registered via the device tree (I'm not quite sure).

Eventually, I found that [QEMU](https://github.com/qemu/qemu/blob/7598971167080a8328a1b8e22425839cb4ccf7b7/hw/virtio/virtio-mmio.c#L213) just passes `-1` to register **0xb0 (SHMLenLow)** and **0xb4 (SHMLenHigh)**, which just works for me now.

shengwen-tw:
embear illustrates a method for [accessing DRM Framebuffer to display an image](https://embear.ch/blog/drm-framebuffer).

The article first shows that to test whether a display is working or not, we can use:
```
$ dd if=/dev/urandom of=/dev/fb0
```

[...]

PR #122, opened 2026-03-19
Implement virtio-input for keyboard and mouse

Mes0903:
This pull request is derived from #34 and #121, implements VirtIO-Input keyboard and mouse devices for semu, enabling host keyboard and mouse input to reach the guest through an SDL-backed window path, in conformance with the [VirtIO specification v1.3](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html).

## VirtIO-Input

- Implements separate VirtIO-Input keyboard and mouse devices
- Uses the VirtIO MMIO transport layer
- Supports:
  - keyboard key press and release events
  - mouse button events
  - absolute pointer motion
  - scroll wheel events
- Implements the VirtIO-Input configuration selectors required by the Linux driver:
  - `VIRTIO_INPUT_CFG_ID_NAME`
  - `VIRTIO_INPUT_CFG_ID_SERIAL`
  - `VIRTIO_INPUT_CFG_ID_DEVIDS`
  - `VIRTIO_INPUT_CFG_PROP_BITS`
  - `VIRTIO_INPUT_CFG_EV_BITS`
  - `VIRTIO_INPUT_CFG_ABS_INFO`
- Handles both `EVENTQ` and `STATUSQ`
  - `EVENTQ` is used for keyboard and mouse event delivery
  - `STATUSQ` is drained and acknowledged so guest LED updates can make forward progress without stalling the queue

### Linux

```sh
sudo apt install libsdl2-dev
```

### macOS

```sh
brew install sdl2
```

## Running semu

```sh
make check
```

If SDL is available, semu should open a window for input testing.

## Basic guest verification

After the guest has booted, verify that the VirtIO-Input devices are present:

```sh
ls /dev/input/event*
cat /proc/bus/input/devices
```

The guest should expose `/dev/input/event*` nodes, and `/proc/bus/input/devices` should report VirtIO input devices.

## Manual end-to-end verification

1. Launch semu:

   ```sh
   make ENABLE_INPUT_DEBUG=1 check
   ```

   If SDL is available, semu should open a window for input testing. With `ENABLE_INPUT_DEBUG=1`, semu also prints the host input events translated by the SDL backend, including:
   
   - keyboard press/release events
   - mouse button press/release events
   - mouse motion coordinates
   - wheel deltas
   - queue-full drop messages

2. In the guest, verify that the VirtIO-Input devices are present:

   ```sh
   ls /dev/input/event*
   cat /proc/bus/input/devices
   ```

3. Optionally inspect the guest event stream directly:

   ```sh
   hexdump /dev/input/event1
   ```

4. Focus the SDL window and:
   - type keys
   - click mouse buttons
   - move the mouse
   - scroll the wheel

You should see the translated host-side events in semu's debug output, and new records should appear in the guest input event stream.

## Automated testing

- Adds `.ci/test-vinput.sh` as a smoke test
- Verifies:
  - `/dev/input/event*` exists
  - `/proc/bus/input/devices` reports VirtIO input devices
- Runs in GitHub Actions on both Linux and macOS
- Uses `SDL_VIDEODRIVER=offscreen` in CI so the guest-visible device
  enumeration path can be validated without requiring a visible desktop
  window

> The current automated test covers device enumeration in the guest. It does not yet validate the actual host->guest input injection path, which is still verified manually.

---
## Summary by cubic
Implements `virtio-input` keyboard and mouse over VirtIO‑MMIO with an SDL‑backed window so host input reaches guest evdev. The window runs on the main thread; the emulator runs in a worker and uses a coalesced wake pipe plus bounded per‑device SPSC queues for fast IRQs and clean shutdown (including gdbstub and window close).

- **New Features**
  - Separate keyboard and mouse; key press/release (`EV_REP` advertised, host autorepeat dropped), mouse buttons, relative motion (`REL_X`/`REL_Y`), and wheel (`REL_WHEEL`/`REL_HWHEEL`). Click window to grab; Ctrl+Alt+G releases.
  - Linux-required config selectors (`ID_NAME`, `ID_SERIAL`, `ID_DEVIDS`, `PROP_BITS`, `EV_BITS`, `ABS_INFO`); unknown selectors return `size=0`. Drains `STATUSQ` for LED updates (no host LED control).
  - Capabilities match the backend: keyboard `EV_KEY` bitmap from the SDL key map, `INPUT_PROP_POINTER`, and relative axes only.
  - SDL/main thread translates host input into bounded per-device SPSC queues; the emulator thread drains events, validates virtqueue bounds/descriptors, coalesces wakeups via a self‑pipe, and clears per‑device queues on reset. Window‑close requests stop the emulator cleanly.
  - Optional input debug logging via `ENABLE_INPUT_DEBUG=1` prints injected events and queue‑drop messages.

- **Dependencies**
  - Build gated on `SDL2`; if `sdl2-config` is missing, `virtio-input` is disabled. CI installs `SDL2`, exports `SDL_VIDEODRIVER=offscreen`, adds `.ci/test-vinput.sh` on Linux/macOS, and extends macOS timeouts.
  - Guest Linux enables `CONFIG_INPUT_EVDEV` and `CONFIG_VIRTIO_INPUT`; the device tree adds two VirtIO‑MMIO input nodes with separate IRQs. CI forces fresh `Image`/`rootfs.cpio` downloads to avoid stale artifacts.

<sup>Written for commit 5c24de0373a7b6cae74d6d532ee3813305536e0a. Summary will update on new commits.</sup>


Mes0903:
@cubic-dev-ai

virtio-input-event.c
```diff
@@ -0,0 +1,343 @@
...
+    uint32_t tail = __atomic_load_n(&window_event_tail, __ATOMIC_ACQUIRE);
+    uint32_t next = (head + 1U) & WINDOW_EVENT_QUEUE_MASK;
+
+    /* Keep the producer non-blocking. If the queue is full, the newest event
+     * is dropped. This remains intentionally lossy even for key/button
+     * events, which means a sustained overflow can lose a release edge. We
+     * keep that tradeoff explicit here rather than synthesizing corrective
+     * events.
+     */
+    if (next == tail)
```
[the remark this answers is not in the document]
Mes0903: As we said, we are going to keep that tradeoff explicit here rather than synthesizing corrective events.

Mes0903:
The macOS netdev CI test timed out again. From the log, it appears to be stuck at `Starting klogd: OK`. This makes me suspect that there may be a condition in the macOS environment that can cause the emulator to hang. For now, I have extended the timeout to two hours so I can determine whether this is really just a timeout issue, or whether there is an actual bug that only occurs on macOS.

Mes0903:
While testing #123 on my own remote, I found that the sound test also experienced timeouts. At the same time, this PR has also shown timeout behavior. These timeouts have occurred in both the netdev and sound tests, but they appear to be random and are difficult to reproduce.

After extending the timeout to two hours, I found that the jobs were not consistently hanging, but rather sometimes just running much more slowly. The difference usually starts to appear only after the init process has been launched, and the gap between the fast and slow cases can be quite large.

For example, the [virtio-input test](https://github.com/sysprog21/semu/actions/runs/23490476364/job/68357052792?pr=122) just took 31 minutes, whereas it normally only takes about 5 minutes, such as in [this run](https://github.com/sysprog21/semu/actions/runs/23488424530/job/68349877424?pr=122). The codebase did not change at all between these two runs. Taking the `EXT4-fs (vda): mounted filesystem` message as an observation point, the former recorded a timestamp of `[  108.175012]`, while the latter recorded `[   22.946067]`, which is far smaller.

Based on this, I believe the problem is most likely with the GitHub runner. The fact that #123 also experienced timeouts is a key reason for this conclusion, because #123 does not change the overall architecture at all, it only fixes a value inside virtio-fs, which should not have any impact on execution time.

mk/external.mk
```diff
@@ -3,12 +3,12 @@
 #   _DATA : the file to be read by specific executable.
 #   _DATA_SHA1 : the checksum of the content in _DATA
 
-COMMON_URL = https://github.com/sysprog21/semu/raw/blob
+COMMON_URL = https://github.com/Mes0903/semu/raw/blob
```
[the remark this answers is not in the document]
Mes0903: It's an temporary commit just for CI testing in this PR, so Im gonna ignore this issue.

.ci/common.sh
```diff
@@ -45,7 +48,7 @@ get_timeout() {
 
     case "${OS_TYPE}" in
         Darwin)
-            echo "900"
+            echo "7200"
```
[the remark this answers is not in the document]
Mes0903: It's an temporary commit for CI testing, Im gonna ignore this issue.

.gitignore
```diff
@@ -11,3 +11,10 @@ rootfs.cpio
 
 # intermediate
 riscv-harts.dtsi
+.smp_stamp
```
shengwen-tw, line 14: Create a dedicated commit to update the `.gitignore` file.

virtio-input-codes.h
```diff
@@ -0,0 +1,205 @@
+#pragma once
+
+/*
+ * SEMU Input Event Codes
+ *
+ * Input event type and code definitions used by the SEMU virtual input
+ * subsystem. The numeric values are chosen to match the Linux evdev ABI so
+ * that a guest kernel can consume them directly without translation.
+ *
+ * Reference: Linux kernel include/uapi/linux/input-event-codes.h
```
shengwen-tw: I believe @jserv suggested removing the reference to the Linux kernel code. Let’s save the link in #32  in case we need it later.

virtio-input.c
```diff
@@ -0,0 +1,926 @@
...
+    VINPUT_MOUSE_NAME,
+};
+
+static inline void bitmap_set_bit(uint8_t *map, unsigned long bit)
+{
+    /* Each byte holds 8 bits. Index into the byte with bit/8,
+     * then set the corresponding bit within that byte with 1 << (bit%8).
+     */
+    map[bit / 8] |= (uint8_t) (1U << (bit % 8));
+}
```
shengwen-tw, line 123: bitmap functions can be reused so I suggest putting them in `common.h`

virtio-input-event.c
```diff
@@ -0,0 +1,371 @@
...
+
+/* Queue storage plus producer/consumer cursors. The queue stays entirely on
+ * the host side so SDL never touches guest-facing virtio-input state directly.
+ */
+static virtio_input_host_event_t
+    virtio_input_event_queue[VIRTIO_INPUT_EVENT_QUEUE_SIZE];
+static uint32_t virtio_input_event_head;
+static uint32_t virtio_input_event_tail;
+static bool virtio_input_event_wake_pending;
+#define DEF_KEY_MAP(_sdl_scancode, _linux_key)                 \
```
shengwen-tw: Move the `#define` above the static variable declarations.

main.c
```diff
@@ -197,6 +224,22 @@ static inline void emu_tick_peripherals(emu_state_t *emu)
...
+        /* The empty path is common during CI and boot workloads, so only
+         * drain the host-side queue after the window thread has published
+         * pending work for the emulator thread.
+         */
+        if (virtio_input_events_maybe_pending())
+            virtio_input_drain_input_events();
+
+        if (virtio_input_irq_pending(&emu->vkeyboard))
+            emu_update_vinput_keyboard_interrupts(vm);
+        if (virtio_input_irq_pending(&emu->vmouse))
```
shengwen-tw, line 249: Insert an empty line here.

main.c
```diff
@@ -197,6 +224,22 @@ static inline void emu_tick_peripherals(emu_state_t *emu)
...
+            emu_update_vinput_mouse_interrupts(vm);
+        /* A closed window is treated like a frontend shutdown request. */
```
shengwen-tw, line 252: Ditto.

main.c
```diff
@@ -1337,6 +1447,19 @@ static int semu_run(emu_state_t *emu)
...
             }
 
+#if SEMU_HAS(VIRTIOINPUT)
+            /* Drain the wake pipe if it fired. emu_tick_peripherals will
+             * then drain queued backend work and check
+             * g_window.window_is_closed().
+             */
+            if (wake_pfd_index >= 0 &&
+                (pfds[wake_pfd_index].revents & POLLIN)) {
+                char buf[16];
```
shengwen-tw: Can we avoid hard-coded `16`?

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+        if (virtio_input_event_rearm_wake())
+            break;
+    }
+}
+
+static void virtio_input_properties(int dev_id)
+{
+    struct virtio_input_config *cfg = &vinput_dev[dev_id].cfg;
+
+    memset(cfg->u.bitmap, 0, 128);
```
shengwen-tw: Let's define a macro for the bitmap size.

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+        cfg->size = 1;
+        break;
+    }
+}
+
+static void virtio_keyboard_support_events(int dev_id, uint8_t event)
+{
+    struct virtio_input_config *cfg = &vinput_dev[dev_id].cfg;
+
+    memset(cfg->u.bitmap, 0, 128);
```
shengwen-tw: Ditto.

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+
+    switch (event) {
+    case SEMU_EV_KEY:
+        /* Only advertise key codes that key_map[] actually generates. */
+        cfg->size =
+            (uint8_t) virtio_input_fill_ev_key_bitmap(cfg->u.bitmap, 128);
```
shengwen-tw: Ditto.

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+        break;
+    default:
+        cfg->size = 0;
+    }
+}
+
+static void virtio_mouse_support_events(int dev_id, uint8_t event)
+{
+    struct virtio_input_config *cfg = &vinput_dev[dev_id].cfg;
+    memset(cfg->u.bitmap, 0, 128);
```
shengwen-tw: Ditto.

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+     *
+     * Check virtinput_cfg_bits() function in linux kernel also.
+     */
+    switch (event) {
+    case SEMU_EV_KEY:
+        bitmap_set_bit(cfg->u.bitmap, SEMU_BTN_LEFT);
+        bitmap_set_bit(cfg->u.bitmap, SEMU_BTN_RIGHT);
+        bitmap_set_bit(cfg->u.bitmap, SEMU_BTN_MIDDLE);
+        /* BTN_LEFT=0x110=272, BTN_MIDDLE=0x112=274 → byte 34 is highest */
+        cfg->size = 35;
```
shengwen-tw: Can we implement a helper to check the highest bit position?

virtio-input.c
```diff
@@ -0,0 +1,919 @@
...
+        break;
+    case SEMU_EV_REL:
+        bitmap_set_bit(cfg->u.bitmap, SEMU_REL_X);
+        bitmap_set_bit(cfg->u.bitmap, SEMU_REL_Y);
+        bitmap_set_bit(cfg->u.bitmap, SEMU_REL_HWHEEL);
+        bitmap_set_bit(cfg->u.bitmap, SEMU_REL_WHEEL);
+        /* REL_WHEEL=8 is the highest advertised relative axis, so the guest
+         * must read two bytes.
+         */
+        cfg->size = 2;
```
shengwen-tw: Ditto.

main.c
Mes0903: I found that this should just read a single byte.

This pipe is only used as a wake-up mechanism to make the emulator thread break out of `poll(-1)` when backend work arrives or when the SDL window is shutting down, it is not carrying any meaningful payload.

The important part here is that we should not leave behind unread wake bytes that would cause an unnecessary immediate wake on the next `poll()`. For the virtio-input path, that is already handled by the `wake_pending` boolean in `virtio-input-event.c`, which coalesces backend wakeups and guarantees that at most one notification byte from virtio-input can be queued in the pipe before the emulator thread drains pending work. If shutdown ends up writing multiple bytes, that is still fine: once we observe the first shutdown wakeup, we are going to notice `g_window.window_is_closed()` and stop anyway, so fully draining repeated shutdown writes is not necessary for correctness.

I will update this to read a single byte and add a more explicit comment describing that reasoning.

jserv (changes requested):
Rebase latest 'master' branch

jserv (changes requested):
Per `CONTRIBUTING.md`, you must indent source files with clang-format-20.

.github/workflows/main.yml
```diff
@@ -103,11 +107,11 @@ jobs:
     - name: default build
       run: make
       shell: bash
-      timeout-minutes: 20
+      timeout-minutes: 120
```
jserv: Is it needed?

virtio-input.c
```diff
@@ -0,0 +1,916 @@
+#include <assert.h>
+#include <errno.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "common.h"
```
jserv: Don't include `common.h` as `Makefiles` does include it.

.github/workflows/main.yml
Mes0903: No. This was included in the "DO NOT MERGE" commit to suppress the time out of CI for testing purposes.

main.c
```diff
@@ -1549,15 +1697,36 @@ static int semu_run_debug(emu_state_t *emu)
...
+    if (emu->debug)
+        ret = semu_run_debug(emu);
+    else
+        ret = semu_run(emu);
+
+    /* Unblock window_main_loop() on the main thread so it can return */
+    if (g_window.window_shutdown)
+        g_window.window_shutdown();
+
+    return (void *) (intptr_t) ret;
```
shengwen-tw: Can we just return `(void *) ret`?

main.c
```diff
@@ -1571,10 +1740,43 @@ int main(int argc, char **argv)
...
+        /* window_main_loop() returns either because the user closed the window
+         * (SDL_QUIT) or because the emulator called window_shutdown().
+         * emu_tick_peripherals() picks up g_window.window_is_closed() and
+         * sets emu->stopped, so no direct write to emu.stopped is needed
+         * here.
+         */
+
+        /* Wait for emulator thread to finish */
+        pthread_join(emu_thread, &thread_ret);
+        ret = (int) (intptr_t) thread_ret;
```
shengwen-tw: Same here.

common.h
```diff
@@ -17,6 +19,20 @@ static inline int ilog2(int x)
...
+    /* Each byte holds 8 bits. Index into the byte with bit/8,
+     * then set the corresponding bit within that byte with 1 << (bit%8).
+     */
+    map[bit / 8] |= (uint8_t) (1U << (bit % 8));
+}
+
+static inline unsigned long bitmap_size_for_highest_bit(
+    unsigned long highest_bit)
+{
+    return (highest_bit / 8) + 1;
```
shengwen-tw: Can we avoid assigning highest_bit manually here?
Given the bitmap size, the function should be able to detect the highest set bit automatically instead of relying on the caller to provide it.

shengwen-tw:
Please update `README.md` to reflect the features supported by virtio-input.

main.c
Mes0903: Currently, the return value is kept as a local variable, so this code uses the pthread start routine's `void *` return value to carry the integer status back to the joining thread.

The conversion here implementation-defined, the conversion path is:

```
int -> intptr_t -> void *    // in emu_thread_func()
void * -> intptr_t -> int    // in main()
```

The reason for doing this is that the emulator return value is not currently stored in `emu_state_t`. Since `pthread_create()` only gives the thread routine one `void *` argument, and that argument is already used for `emu`, `pthread_join()`'s `retval` is used to propagate the return value of `semu_run()`/`semu_run_debug()`
from `emu_thread_func()`.

Given that, I would prefer not to simplify this to a direct cast such as `(void *)ret` or `(int)thread_ret`. If we keep the return value as a local variable, the explicit cast through `intptr_t` makes the integer/pointer conversion clear
and avoids direct int/pointer size casts.

Alternatively, we could store the emulator return value in `emu_state_t` instead. That may be cleaner overall, because `pthread_join()` would then only be used for synchronization and the thread routine could simply return `NULL`.

jserv (changes requested):
Rebase latest `master` branch which results in faster executions.

main.c
Mes0903: Solved by move the return variable into `emu_state_t`, now it will modify `emu->exit_code` directly.

main.c
Mes0903: Solved by move the return variable into `emu_state_t`, now it will modify `emu->exit_code` directly.

shengwen-tw (approved):

jserv (changes requested):
Drop "DO NOT MERGE" commits. Then, I will merge this PR.

Mes0903:
The current CI failure is expected because the Image on the main blob branch is still the old one, and CI does not rebuild the kernel Image.

I've manually rebuilt the kernel locally with the current config and tested it by hand. The virtio-input functionality works correctly with the rebuilt Image.

jserv:

PR #132, opened 2026-04-25
Fix virtio-snd deadlock and concurrency bugs

### LINKED ISSUE #130: Sound Playback Hangs without Any Playing Sound
After commit `e9a64b7`, the sound playback hangs without any playing sound when trying to play sound:

```
# aplay /usr/share/sounds/alsa/Front_Center.wav 
Playing WAVE '/usr/share/sounds/alsa/Front_Center.wav' : Signed 16 bit Little Endian, Rate 48000 Hz, Mono

... wait in here indefinitely 
```

The reason is that there occurs a deadlock in `virtio-snd.c`:

```c
// in virtio_snd_reg_write()
case VSND_QUEUE_TX:
                pthread_mutex_lock(&virtio_snd_mutex);
                tx_ev_notify++;
                pthread_cond_signal(&virtio_snd_tx_cond);
                pthread_mutex_unlock(&virtio_snd_mutex);
                break;

// in *func()
for (;;) { 
        pthread_mutex_lock(&virtio_snd_mutex);
        while (tx_ev_notify <= 0)
            pthread_cond_wait(&virtio_snd_tx_cond, &virtio_snd_mutex);
        
        tx_ev_notify--;
        virtio_queue_notify_handler(vsnd, 2);

        pthread_mutex_unlock(&virtio_snd_mutex);
    }
```

jserv:
This resolves the sound playback hang where aplay blocks indefinitely. The root cause: commit 18a3dcb correctly added mutex protection around the TX condvar signal, but the TX worker thread held virtio_snd_mutex during the entire virtio_queue_notify_handler() call, which can block inside __virtio_snd_frame_enqueue() waiting for the PortAudio callback to drain frames. Release the mutex before the handler call since it only needs to guard tx_ev_notify.

Close #130

---
## Summary by cubic
Fixes a virtio-snd deadlock that caused aplay to hang. Releases the TX mutex before queue notify and hardens concurrency with atomics and cancellable waits to prevent stalls and missed interrupts. Fixes #130.

- **Bug Fixes**
  - Released the TX mutex before calling queue notify in the TX thread to avoid blocking in frame enqueue when PortAudio is draining.
  - Hardened queue condvar handling: signal readable while holding the queue lock; make enqueue/dequeue waits cancellable via a new releasing flag; broadcast on release; zero-fill dequeue on release.
  - Switched `Status`/`InterruptStatus` to atomic ops with proper ordering; publish used-ring writes before raising IRQ; ACK via atomic AND; `main.c` now loads `InterruptStatus` atomically.
  - Reduced notify handler races by loading `Status` once, returning early when the device needs reset, and only proceeding when `DRIVER_OK` is set and the queue is ready; otherwise set fail.

<sup>Written for commit fbcca5c17d08dcc34e486fa40bca51252a5a6a7c. Summary will update on new commits.</sup>

### DIFF
--- a/main.c
+++ b/main.c
@@ -159,7 +159,7 @@ static void emu_update_swi_interrupt(hart_t *hart)
 static void emu_update_vsnd_interrupts(vm_t *vm)
 {
     emu_state_t *data = PRIV(vm->hart[0]);
-    if (data->vsnd.InterruptStatus)
+    if (__atomic_load_n(&data->vsnd.InterruptStatus, __ATOMIC_ACQUIRE))
         data->plic.active |= IRQ_VSND_BIT;
     else
         data->plic.active &= ~IRQ_VSND_BIT;
@@ -227,7 +227,7 @@ static inline void emu_tick_peripherals(emu_state_t *emu)
 #endif
 
 #if SEMU_HAS(VIRTIOSND)
-        if (emu->vsnd.InterruptStatus)
+        if (__atomic_load_n(&emu->vsnd.InterruptStatus, __ATOMIC_ACQUIRE))
             emu_update_vsnd_interrupts(vm);
 #endif
 
--- a/virtio-snd.c
+++ b/virtio-snd.c
@@ -270,6 +270,7 @@ typedef struct {
 typedef struct {
     pthread_cond_t readable, writable;
     int buf_ev_notify;
+    int releasing;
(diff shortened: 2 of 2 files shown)


Cuda-Chen (approved):
Code LGTM.

I have confirmed the change does not hang the audio playing.

PR #133, opened 2026-04-27
Implement virtio-gpu 2D

Mes0903:
This pull request implements a minimal VirtIO-GPU 2D device for semu. The device uses the VirtIO MMIO transport and provides DRM/KMS support for Linux to bind the `virtio_gpu` driver and expose `/dev/dri/card0`. The software backend supports 2D resource management, scanout updates, and cursor presentation through the SDL window backend.

When the SDL window grabs the pointer, press `Ctrl+Alt+G` to release the mouse cursor back to the host.

## VirtIO-GPU

- Adds a VirtIO-GPU MMIO device.
- Implements the basic 2D command path:
  - display information and EDID queries
  - 2D resource creation and destruction
  - backing attachment and detachment
  - guest-to-host framebuffer transfer
  - scanout configuration and resource flushing
  - cursor set, move, and clear operations
- Adds a software 2D backend.
- Presents scanout and cursor updates through the SDL window backend.
- Adds a virtio-gpu CI smoke test that checks Linux driver binding,
  `/dev/dri/card0`, and the DirectFB2 DRM/KMS path.

## Tested workloads

The implementation has been tested with:

- Linux `virtio_gpu`
- libdrm test tools such as `modetest`
- DirectFB2 using its DRM/KMS backend
- X11
- `kmscube`
- `glmark2`
- `glxgears`

## Build and test

Install SDL2 development headers if they are not already present:

```sh
sudo apt install libsdl2-dev libsdl2-2.0-0
```

Build the emulator:

```sh
make
```

Run the virtio-gpu CI smoke test:

```sh
.ci/test-gpu.sh
```

To rebuild the default guest artifacts locally:

```sh
./scripts/build-image.sh --all
```

To build the DirectFB2 test tool disk used by the GPU smoke test:

```sh
./scripts/build-image.sh --directfb2-test
```

For the temporary X11-enabled test image used during review and CI testing:

```sh
./scripts/build-image.sh --x11 --directfb2-test
```

For a full manual GPU test run, build or download the kernel and tool disk, then launch semu with the tool disk as the guest root filesystem:

```sh
make semu minimal.dtb Image tool-ext4.img
./semu -k Image -c 1 -b minimal.dtb -d tool-ext4.img
```

## Additional test scaffolding

This PR currently carries two `DO NOT MERGE` commits to make review and CI easier while the final guest artifact flow is being prepared:

- `DO NOT MERGE: Add X11 image support for CI testing` adds an optional X11-enabled Buildroot configuration. The normal `rootfs.cpio` and `ext4.img` stay on the default Buildroot configuration, while the heavier X11 payload is built into `tool-ext4.img` for development and CI coverage.
- `DO NOT MERGE: Cache PR-built guest artifacts` keeps PR-built `Image`/`rootfs.cpio`/`tool-ext4.img` artifacts cached by the live guest-input hash, then re-uploads them as the normal `prebuilt-pr` workflow artifact for downstream CI jobs. It also prevents restored PR artifacts from falling back to the upstream release download path before `tool-ext4.img` exists there.

These commits are test scaffolding only. They are expected to be dropped or replaced before merge and should not be treated as part of the final user-facing build flow. The temporary CI/cache/X11 adjustments are kept in `DO NOT MERGE` commits and are not intended to land in `master`.

### Check DRM/KMS binding

Inside the guest:

```sh
ls -l /dev/dri/
modetest -M virtio_gpu
```

### Run modetest

Use the connector ID reported by `modetest -M virtio_gpu`:

```sh
modetest -M virtio_gpu -s <connector-id>:1024x768
```

A test image should appear in the SDL window.

If you see `Invalid argument`, re-check the connector ID. If you see `Permission denied` or the DRM device is busy, stop the process that currently owns `/dev/dri/card0`:

```sh
ps | grep -E "Xorg|weston|kmscube|DirectFB"
kill <PID>
```

### Run DirectFB2

If the DirectFB2 test payload is present in the guest disk:

```sh
. /root/local-env.sh
df_drivertest
```

Other DirectFB2 examples can also be run from the same environment, such as:

```sh
df_fire
df_matrix
df_window
df_input
df_layers
```

The installed DirectFB2 examples come from the upstream `DirectFB-examples` project and can be listed in the guest with:

```sh
ls /usr/local/bin/df_*
```

If X11 is running and owns `/dev/dri/card0`, stop Xorg first before running DirectFB2:

```sh
ps | grep Xorg
kill <PID>
```

### Run kmscube

If the test image includes `kmscube`, stop X11 first if it is running, then run:

```sh
kmscube
```

### Run X11 workloads

If using the temporary X11-enabled test image from the `DO NOT MERGE` scaffolding, start X11 inside the guest:

```sh
startx
```

After X11 is running, launch the test programs from the guest:

```sh
glmark2
glxgears
```


.ci/test-gpu.sh
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
```
jserv, line 3: For consistency, rename to `.ci/test-gpu.sh`

scripts/build-image.sh
```diff
@@ -133,11 +165,73 @@ while [[ $# -gt 0 ]]; do
...
+{
+    export PATH=$PATH:$PWD/buildroot/output/host/bin
+    export BUILDROOT_OUT=$PWD/buildroot/output/
+    export DIRECTFB_STAGE=$PWD/directfb
+    mkdir -p directfb
+
+    # Build DirectFB2
+    if [ ! -d DirectFB2 ]; then
+        echo "Cloning DirectFB2..."
+        ASSERT git clone https://github.com/directfb2/DirectFB2.git
```
jserv: Use shorter form. i.e., `https://github.com/directfb2/DirectFB2`

mk/external.mk
```diff
@@ -6,15 +6,20 @@
 COMMON_URL = https://github.com/sysprog21/semu/raw/blob
 
 # kernel
-KERNEL_DATA_URL = $(COMMON_URL)/Image.bz2
+KERNEL_DATA_URL = https://github.com/Mes0903/semu/raw/blob/Image.bz2
```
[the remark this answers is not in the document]
Mes0903: This is introduce by a temporary commit "DO NOT MERGE" for CI testing, so Im going to ignore this one

scripts/build-image.sh
```diff
@@ -86,12 +106,14 @@ function do_linux
 
 function show_help {
     cat << EOF
-Usage: $0 [--buildroot] [--linux] [--all] [--external-root] [--clean-build] [--help]
+Usage: $0 [--buildroot] [--x11] [--linux] [--directfb-test] [--all] [--external-root] [--clean-build] [--help]
 
 Options:
   --buildroot         Build Buildroot rootfs
+  --x11               Build Buildroot with X11
+  --directfb-test     Build an ext4 guest disk with the DirectFB2 test payload
```
shengwen-tw: Since `DirectFB2` is derived from the original `DirectFB` project, I think it would be better to name the option `--directfb2-test`.

Reference:
[DirectFB2](https://github.com/directfb2)
>"DirectFB2 is a fork of DirectFB whose purpose is to preserve and maintain the DirectFB graphics backend, particularly for use on embedded systems."

virtio-gpu.h
```diff
@@ -0,0 +1,387 @@
...
+ *
+ * Supporting that path would require accepting a longer descriptor chain and
+ * auditing every handler that indexes 'vq_desc[]'. Longer chains are rejected.
+ * The current response-descriptor lookup is also part of this fixed-shape
+ * parser: it scans the zero-initialized 3-entry array, not an arbitrary
+ * guest-provided scatter-gather chain.
+ *
+ * TODO: Support generic descriptor-chain parsing.
+ */
+#define VIRTIO_GPU_MAX_DESC 3
```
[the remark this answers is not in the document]
Mes0903: Currently, Linux uses the following declaration in `virtio_gpu_queue_fenced_ctrl_buffer()`:

```c
struct scatterlist *sgs[3], vcmd, vout, vresp;
```

These three top-level segments are:

* `vcmd`: control request
* `vout`: optional command data, such as `RESOURCE_ATTACH_BACKING` entries
* `vresp`: optional response

However, this does not guarantee that the request uses at most three virtqueue descriptors. Linux allocates the backing-entry array in `virtio_gpu_object_shmem_init()` using `kvmalloc_objs()`. Since `kvmalloc` may fall back to `vmalloc`, when `vbuf->data_buf` is a vmalloc address, `virtio_gpu_queue_fenced_ctrl_buffer()` detects this with `is_vmalloc_addr()` and switches to `vmalloc_to_sgt()`, expanding the original `vout` into multiple scatter-gather entries.

The current semu virtio-gpu implementation does not have a negotiated limit like virtio-blk’s `seg_max = 1` that can require Linux to keep `vout` as a single descriptor. Therefore, `VIRTIO_GPU_MAX_DESC == 3` is only the simplified path currently supported by semu, not the complete descriptor-chain behavior of Linux virtio-gpu.

In the current test environment, the framebuffer backing-entry array is relatively small and usually goes through the contiguous `kmalloc` path, so this limitation is unlikely to be triggered. However, for larger resources or under guest kernel memory fragmentation, it is indeed possible for `kvmalloc` to fall back to `vmalloc`, in which case Linux may submit a descriptor chain with more than three descriptors.

That said, since the current virtio-gpu implementation is intended to be a minimal 2D model, I think this limitation can be deferred for now. Trying to fix it would require a non-trivial refactoring of the current virtqueue handling model, rather than a small local change in virtio-gpu. Similar simplifications also exist in the current virtio-fs, virtio-blk, and virtio-snd implementations. 

The virtio-blk case is slightly better justified because its data segment is effectively constrained to a single SG entry through feature negotiation, so the current three-descriptor assumption in `virtio-blk.c` has negotiated support. In contrast, virtio-gpu and virtio-fs do not currently have an equivalent negotiated limit, so their behavior relies on a weaker assumption.

For this reason, I think the better long-term direction is to open a separate issue and factor the virtqueue logic into a shared implementation that can be reused by all virtio devices, instead of having each device redefine its own virtqueue traversal and descriptor-handling flow. The current simplification has also been documented in the code comments, including the rationale behind this decision.

README.md
```diff
@@ -125,12 +129,13 @@ This command invokes the underlying script: `scripts/build-image.sh`, which also
 ### Script Usage
 
 ```
-./scripts/build-image.sh [--buildroot] [--linux] [--all] [--external-root] [--clean-build] [--help]
+./scripts/build-image.sh [--buildroot] [--linux] [--directfb2-test] [--all] [--external-root] [--clean-build] [--help]
```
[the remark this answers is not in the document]
Mes0903: This is introduce by a temporary commit "DO NOT MERGE" for CI testing, so Im going to ignore this one

mk/external.mk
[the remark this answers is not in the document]
Mes0903: This is introduce by a temporary commit "DO NOT MERGE" for CI testing, so Im going to ignore this one

jserv (changes requested):
Rebase latest 'master' branch, as deployment and build system are refined.

Mes0903:
@jserv Since master now rebuilds PR prebuilts from source and publishes only `Image/rootfs.cpio`, should the DirectFB2 GPU test program become part of the normal Buildroot rootfs, or should `test-gpu` build/inject a separate GPU-only ext4 image during CI?

jserv:
> Since master now rebuilds PR prebuilts from source and publishes only `Image/rootfs.cpio`, should the DirectFB2 GPU test program become part of the normal Buildroot rootfs, or should `test-gpu` build/inject a separate GPU-only ext4 image during CI?

Both are doable. You can even propose alternatives. The only consideration is to minimize kernel/rootfs images when possible.

configs/buildroot.config
```diff
@@ -39,6 +39,8 @@ BR2_FORTIFY_SOURCE_1=y
 BR2_PACKAGE_ALSA_UTILS=y
 BR2_PACKAGE_ALSA_UTILS_APLAY=y
 BR2_PACKAGE_ALSA_UTILS_SPEAKER_TEST=y
+BR2_PACKAGE_LIBDRM=y
+BR2_PACKAGE_LIBDRM_INSTALL_TESTS=y
```
shengwen-tw: Let’s disable the `LibDRM tests` since they’re minor and would increase the image size.

Mes0903:
I decided to keep the normal kernel/rootfs path small and keep the GPU test payload out of the default Buildroot rootfs.

The current direction is to keep `rootfs.cpio` and the default `ext4.img` built from the normal Buildroot config, then build a separate `tool-ext4.img` only for GPU-oriented CI tests. `test-gpu.sh` boots that tool disk directly when it needs the DirectFB2 test programs. This keeps the default boot artifacts small, while still letting CI exercise the DRM/KMS path end to end.

For the temporary X11/glxgears validation, the `DO NOT MERGE` commit extends that same tool-image path with `--x11`; it does not put X11 into the normal rootfs. That part is only for PR testing and should not go into the mainline series.

Mes0903:
I also made a few CI-only adjustments after testing the PR prebuilt path.

The PR prebuilt job now caches the generated guest artifacts by the live guest-input hash. When the PR changes Buildroot/Linux/rootfs inputs, the first run still rebuilds Image/rootfs.cpio/tool-ext4.img from source, but later pushes with the same guest inputs can restore those files from cache instead of running Buildroot again.

I also kept the downstream jobs artifact-based: pr-prebuilt-build still uploads Image/rootfs.cpio/tool-ext4.img as prebuilt-pr, so the Linux/macOS test jobs do not need to know whether the files came from a fresh build or from cache.

One related Makefile issue showed up during CI: external artifacts depended on prebuilt.sha1 as a normal prerequisite, and prebuilt.sha1 is refreshed via FORCE. That made make check treat already-restored PR artifacts as stale and try to download tool-ext4.img.bz2 from the upstream release, where it does not exist yet. I changed that dependency to order-only, so the manifest is still available when an artifact has to be downloaded, but existing PR-built artifacts are not redownloaded just because the manifest was refreshed.

Finally, test-gpu.sh now checks whether Image/rootfs.cpio/tool-ext4.img are already present before invoking make for them. This lets the workflow artifact path work correctly and avoids accidentally falling back to the release download path.

All of these CI adjustments are currently kept in `DO NOT MERGE` commits, so they are only for validating this PR and are not intended to land in master.

Mes0903:
For convenience when manually testing the PR, I have also uploaded the prebuilt files to my own blob branch: [https://github.com/Mes0903/semu/tree/blob](https://github.com/Mes0903/semu/tree/blob)

main.c
```diff
@@ -993,25 +1024,46 @@ static int semu_init(emu_state_t *emu, int argc, char **argv)
...
-        fcntl(emu->wake_fd[1], F_SETFL, O_NONBLOCK);
+        int flags = fcntl(emu->wake_fd[1], F_GETFL, 0);
+        if (flags < 0 ||
+            fcntl(emu->wake_fd[1], F_SETFL, flags | O_NONBLOCK) < 0) {
+            perror("fcntl");
+            close(emu->wake_fd[0]);
+            close(emu->wake_fd[1]);
+            emu->wake_fd[0] = emu->wake_fd[1] = -1;
+            g_window.window_cleanup();
+            return 2;
```
shengwen-tw: Use POSIX exit status macro instead of a number.

main.c
```diff
@@ -993,25 +1024,46 @@ static int semu_init(emu_state_t *emu, int argc, char **argv)
...
+
+        /* Make the write end non-blocking so 'window_shutdown_sw()' never
+         * stalls. The read end remains blocking because 'semu_run()' reads it
+         * only after 'poll()' reports 'POLLIN' on the same emulator thread.
          */
-        fcntl(emu->wake_fd[1], F_SETFL, O_NONBLOCK);
+        int flags = fcntl(emu->wake_fd[1], F_GETFL, 0);
+        if (flags < 0 ||
+            fcntl(emu->wake_fd[1], F_SETFL, flags | O_NONBLOCK) < 0) {
+            perror("fcntl");
```
shengwen-tw: Please make the error message more explicit.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+                            10000U;
+        if (pixel_clock_10khz > 0xffffU)
+            pixel_clock_10khz = 0xffffU;
+    }
+
+    /* Clamp fields to the bit widths defined by Table 3.21:
+     * active/blanking/image-size fields are 12-bit, horizontal sync fields are
+     * 10-bit, and vertical sync fields are 6-bit.
+     */
+    if (width > 4095U)
```
shengwen-tw: Add a saturation/bounding macro to avoid repeating if pattern.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+    /* Check EDID 1.4 Section 3.10.2: detailed timing descriptor layout. */
+    uint32_t h_blank;           /* Horizontal blanking pixels. */
+    uint32_t h_front;           /* Horizontal front porch pixels. */
+    uint32_t h_sync;            /* Horizontal sync pulse width. */
+    uint32_t v_blank;           /* Vertical blanking lines. */
+    uint32_t v_front;           /* Vertical front porch lines. */
+    uint32_t v_sync;            /* Vertical sync pulse width. */
+    uint32_t pixel_clock_10khz; /* Pixel clock in 10 kHz units. */
+
+    if (width == 1024 && height == 768) {
```
shengwen-tw: Avoid hard-coding `width == 1024` and `height == 768`.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+        v_blank = 38;
+        v_front = 3;
+        v_sync = 6;
+    } else {
+        /* Fallback only: scale porch/sync proportions from the VESA DMT
+         * 1024x768@60Hz timing. This is not a CVT/GTF calculation, but keeps
+         * non-1024x768 modes based on a documented timing instead of ad hoc
+         * ratios.
+         */
+        h_blank = ((uint64_t) width * 320U + 1024U / 2U) / 1024U;
```
shengwen-tw: Ditto.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+        if (h_sync == 0)
+            h_sync = 1;
+        if (h_blank <= h_front + h_sync) {
+            /* Keep front porch and sync pulse inside the blanking interval so
+             * the remaining pixels form the back porch.
+             */
+            h_blank = h_front + h_sync + 1U;
+        }
+
+        v_blank = ((uint64_t) height * 38U + 768U / 2U) / 768U;
```
shengwen-tw: Ditto.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+    /* Check EDID 1.4 Section 3.1: base EDID block layout. */
+    if (width == 0)
+        width = SCREEN_WIDTH;
+    if (height == 0)
+        height = SCREEN_HEIGHT;
+
+    uint16_t width_mm = virtio_gpu_edid_pixels_to_mm(width);
+    uint16_t height_mm = virtio_gpu_edid_pixels_to_mm(height);
+
+    memset(edid, 0, 128);
```
shengwen-tw: Define a macro for 128.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+                          (((manufacture[1] - '@') & 0b11111) << 5) |
+                          (((manufacture[2] - '@') & 0b11111) << 0));
+    /* Convert vendor ID to big-endian order */
+    edid[8] = vendor_id >> 8;
+    edid[9] = vendor_id & 0xff;
+
+    /* Check EDID 1.4 Sections 3.4.2 and 3.4.3: product code and serial
+     * number, all zeros if unused.
+     */
+    memset(&edid[10], 0, 6);
```
shengwen-tw: Use `sizeof` if possible.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+void virtio_gpu_init(virtio_gpu_state_t *vgpu)
+{
+    static bool initialized = false;
+
+    if (initialized) {
+        fprintf(stderr,
+                VIRTIO_GPU_LOG_PREFIX
+                "%s(): only one virtio-gpu instance is supported\n",
+                __func__);
+        exit(2);
```
shengwen-tw: Ditto.

virtio-gpu.c
```diff
@@ -0,0 +1,1143 @@
...
+{
+    int scanout_num = PRIV(vgpu)->num_scanouts;
+    if (scanout_num >= VIRTIO_GPU_MAX_SCANOUTS) {
+        /* Registration is init-only today. Return an error instead if scanout
+         * creation becomes dynamic or guest-triggered.
+         */
+        fprintf(stderr,
+                VIRTIO_GPU_LOG_PREFIX "%s(): exceeded scanout maximum number\n",
+                __func__);
+        exit(2);
```
shengwen-tw: Ditto.

shengwen-tw (changes requested):

.ci/publish-prebuilt.sh
```diff
@@ -1,29 +1,32 @@
...
-# splice them into release notes, GITHUB_OUTPUT, or whatever else.
+# Compress the prebuilt Image, rootfs.cpio, and tool-ext4.img in cwd, write a
+# sha1 manifest, hash the input files that define the prebuilt's contents, and
+# print all four sums in KEY=VAL form on stdout so callers can splice them into
+# release notes, GITHUB_OUTPUT, or whatever else.
 #
 # Inputs (in cwd):
 #   Image
 #   rootfs.cpio
+#   tool-ext4.img
```
jserv: The name is confusing. Make it straightforward.

Mes0903: The original reason for the `tool-ext4.img` name was that this image was intended to hold third-party guest-side test tools that are too large or too specialized for the default rootfs. DirectFB2 is the first user of it, but I
was trying not to make the name GPU- or DirectFB2-specific.

How about renaming it to `test-tools.img` instead? The filename would describe the purpose of the disk, while the README can still document that it is an ext4-formatted guest disk image used with `-d test-tools.img`.

Also, do you expect future component-specific tests to each carry their own disk image, or would you prefer a shared test-tools image that additional third-party guest tools can be appended into? My original assumption was the shared image approach, to avoid growing the default rootfs while also avoiding one extra disk artifact per component.

jserv: Don't mention filesystem types in image file name, as it is definitely traceable.

Mes0903: The name has been changed to `test-tools.img` now.

mk/external.mk
```diff
@@ -90,9 +96,12 @@ PREBUILT_INPUTS := \
     configs/linux.config \
     configs/busybox.config \
     configs/buildroot.config \
+    configs/x11.config \
+    configs/riscv-cross-file \
     scripts/build-image.sh \
     scripts/rootfs_ext4.sh \
-    target/init
+    target/init \
+    target/local-env.sh
```
jserv, line 102: Is `local-env.sh` required?

vgpu-display.c
```diff
@@ -0,0 +1,288 @@
...
+        .type = VGPU_DISPLAY_CMD_CURSOR_SET,
+        .scanout_id = scanout_id,
+        .generation =
+            __atomic_load_n(&vgpu_display_cursor_clear[scanout_id].generation,
+                            __ATOMIC_ACQUIRE),
+        .u.cursor_set = {.payload = payload,
+                         .x = x,
+                         .y = y,
+                         .hot_x = hot_x,
+                         .hot_y = hot_y},
```
jserv: Add comma right after `= hot_y` to make clang-format happy.

vgpu-display.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+#include "feature.h"
```
jserv: It is not necessary to include `feature.h` as it is automatically included by build system.

mk/external.mk
Mes0903: It depends on how we want users to invoke the test tools. `local-env.sh` is the guest-side environment hook for tools that are overlaid through the shared `test-tools.img`. Now it adds `/usr/local/bin` to `PATH` and `/usr/local/lib` to `LD_LIBRARY_PATH`. 

The idea is that the default rootfs stays small, while optional test tools are collected in the same `test-tools.img`.  When we add a new test tool later, the image build should stage that tool into `test-tools.img`, and `local-env.sh` should be updated if that tool needs an additional binary or library search path.

With that convention, after booting the VM the user only needs to source `/root/local-env.sh` once, instead of typing full paths such as `/usr/local/bin/df_*` for the overlaid tools.

main.c
```diff
@@ -993,25 +1024,46 @@ static int semu_init(emu_state_t *emu, int argc, char **argv)
...
+    vgpu_display_set_scanout_count(scanout_id + 1U);
+#endif
+
+#if SEMU_HAS(VIRTIOINPUT) || SEMU_HAS(VIRTIOGPU)
+    g_window.window_init(headless, SCREEN_WIDTH, SCREEN_HEIGHT);
 
     emu->wake_fd[0] = emu->wake_fd[1] = -1;
     if (vm->n_hart > 1 && g_window.window_main_loop) {
         if (pipe(emu->wake_fd) < 0) {
             perror("pipe");
```
shengwen-tw: Please add more explicit error message for `perror`.

main.c
```diff
@@ -1583,14 +1660,14 @@ static void semu_run(emu_state_t *emu)
                 perror("poll");
```
shengwen-tw: Ditto.

shengwen-tw (changes requested):

shengwen-tw:
Aside from the CI/test change which I’m not familiar enough, the implementation looks legitimate to me.

README.md
```diff
@@ -19,7 +19,11 @@ A minimalist RISC-V system emulator capable of running Linux the kernel and corr
             - For instance, the following buffer/period size settings on `aplay` has been tested
               with broken and stutter effects yet complete with no any errors: `aplay --buffer-size=32768 --period-size=4096 /usr/share/sounds/alsa/Front_Center.wav`.
     - virtio-input exposes SDL-backed keyboard and mouse devices to the guest.
-      - You can exit the SDL window by pressing Ctrl+A+G
+    - virtio-gpu exposes a minimal 2D DRM/KMS device to the guest. Linux can
+      bind the `virtio_gpu` driver and create `/dev/dri/card0`.
+      - Only 2D scanout is currently supported; 3D, virgl, and blob resources
+        are not implemented yet.
```
jserv, line 25: Add proper FIXME/TODO comments in C source files.

jserv:

PR #140, opened 2026-06-03
Improve build system and CI/CD pipeline

Mes0903:
# Improve build system and CI/CD pipeline

This pull request restructures the guest artifact build flow used by semu.

The main reason for this change is the ordering problem in the old CI/CD flow. When a branch was merged into `master`, the device tests could request the rolling release prebuilt artifacts before the new prebuilt release had been built and published. That meant the tests could boot with stale release artifacts even though the current checkout required new guest artifacts, so the merge-time CI/CD path could fail for reasons unrelated to the emulator changes being tested.

The old PR flow also compared guest artifact inputs against the `master` branch. When a PR changed a guest config, every later push to that PR could trigger another source rebuild, even if the new commit did not touch any guest config or recipe input. This made PR iteration expensive because the decision was tied to branch drift instead of the exact artifact recipe required by the current checkout.

The old flow also made the build and test pipeline depend heavily on GitHub-specific behavior. Cache restore, workflow artifacts, release downloads, publish decisions, and test setup were coupled together in the workflow. This PR redraws that architecture around clearer layers: source-build recipes, artifact resolution, artifact materialization, packaging/publishing, and test consumption. GitHub Actions remains the current execution environment, but the GitHub-specific pieces are now adapters around those layers rather than the place where the artifact policy lives.

Local development stays close to Make’s file-target model. Missing guest artifacts can still be downloaded from the rolling prebuilt release, while source rebuilds are explicit through `make build-artifacts`. This keeps the normal local path predictable and preserves the Buildroot cache workflow developers rely on when iterating on guest configuration.

CI owns the artifact decision model. The workflow computes recipe keys for the current checkout, compares them with restored workflow-local raw artifacts and the rolling release manifest, then materializes a plan as cache reuse, release download, or source build. The resolver decides where artifacts should come from, and the materializer performs that plan.

The rolling release contract is also tightened. Recipe keys continue to decide whether a release artifact matches the current source recipe, while archive checksums verify that downloaded `.bz2` bytes match the published manifest before decompression and stamping. These checks cover different failure modes, so the release manifest now records both archive checksum entries and virtual recipe-key entries.

## Local build flow

`make build-artifacts` replaces the old image-oriented source build entrypoint with an artifact-class based flow. The local CLI selects which classes to build, handles cleanup policy, and delegates the actual source recipes to shared helpers under `scripts/prebuilt/`.

The default rootfs and test-tools rootfs now use separate Buildroot output directories. This keeps optional test payloads such as X11 and DirectFB2 from changing the default `rootfs.cpio` and derived `ext4.img` path. `ext4.img` remains a local derived artifact built from `rootfs.cpio`.

Test-tools payload selection is explicit. Options such as `--x11` and `--directfb2-test` are passed directly to the build command, and DirectFB2 no longer implicitly selects X11.

## CI artifact flow

CI is split into resolution and materialization. The resolver reads the current recipe inputs, restored cache stamps, and release manifest, then emits a plan for each artifact class. The materializer executes that plan without making new policy decisions.

Workflow-local raw artifacts are tracked with `.stamps/prebuilt-local/*.recipe-key`. Those stamps allow restored CI cache entries to be checked against the current checkout before reuse. Test jobs consume either the raw artifact bundle produced by the current workflow run or the rolling prebuilt release through the same composite action.

Publishing is part of the main workflow. After the required tests run against the same materialized artifact set, the publish job verifies the packaged artifacts and updates the rolling `prebuilt` release when the resolver reports release drift.

## Release manifest

`prebuilt.sha1` now records both compressed archive checksums and recipe-key entries. The archive checksum entries verify transport integrity for `Image.bz2`, `rootfs.cpio.bz2`, and `test-tools.img.bz2`. The recipe-key entries describe the source recipe that produced each artifact class.

Local release downloads verify archive bytes before decompression. CI release materialization additionally verifies that the manifest recipe key still matches the resolver plan before stamping the raw artifact as current.

Fork and mirror repositories can bootstrap from source when the rolling release manifest returns HTTP 404. Other release download failures remain visible as release or network state problems.

## Documentation

The README now describes missing guest artifacts as release downloads and source guest rebuilds as an explicit `make build-artifacts` operation.

`docs/build-system-and-ci.md` documents the local build path, CI artifact contract, resolver/materializer model, package and publish flow, and fork/mirror bootstrap behavior.

The networking document is updated to point at the moved device smoke script path.

---
## Summary by cubic
Rebuilt the guest artifact build and CI/CD pipeline around artifact classes with a recipe‑keyed resolver/materializer and verified release archive checksums to prevent stale prebuilts and speed up PRs. Added fast build‑system contract tests, selectable tool hooks for reliable `ext4.img` on macOS, and early validation of default `minimal.dtb` boot options.

- **Refactors**
  - Local builds via `make build-artifacts` backed by `scripts/prebuilt/*`; separate Buildroot outputs for default rootfs vs. test‑tools; explicit test‑tools flags (`--x11`, `--directfb2-test`).
  - CI plans per class (`use-action-cache`/`download-release`/`build`) with stamps in `.stamps/prebuilt-local`; materializer executes the plan; tests get artifacts via `provide-guest-artifacts`; removed the standalone `prebuilt` workflow.
  - Release: `prebuilt.sha1` records `*.bz2` checksums plus `*.recipe-key`; downloads verify archive bytes and expected recipe; package/verify via `.ci/prebuilt/package.sh` and `verify-package.sh`.
  - Tooling: selectable `ROOTFS_CPIO`, `ROOTFS_CP`, `ROOTFS_CHOWN`, and `ROOTFS_FAKEROOT_SHELL` for ext4 builds; added `scripts/check-rootfs-fakeroot.sh`; macOS CI installs Homebrew `cpio` and sets these envs.

- **Migration**
  - Developers: `make check` downloads verified prebuilts; rebuild via `make build-artifacts [image|rootfs|test-tools]` with `--x11`/`--directfb2-test`, `--clean-build`, or `--full-rebuild`. Set `ROOTFS_CPIO` (and related vars) on macOS if needed.
  - Maintainers: rolling `prebuilt` release publishes from the main workflow on `master` after tests; `master` runs aren’t canceled by concurrency.
  - Forks/mirrors: `PREBUILT_URL` defaults to this repository; CI can bootstrap from source when the release manifest returns 404.

<sup>Written for commit 956de182d5a475ea628f1d55ddb8c5fb39c7db0b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/sysprog21/semu/pull/140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Mes0903:
@cubic-dev-ai

Cuda-Chen (approved):

main.c
```diff
@@ -841,8 +840,53 @@ static bool uses_default_minimal_dtb(const char *dtb_file)
...
+
+    if (!disk_file) {
+        fprintf(stderr,
+                "warning: EXTERNAL_ROOT build expects -d <disk-image>; "
+                "without it the kernel will hang in rootwait.\n");
+    }
+#else
+    if (!initrd_file) {
+        fprintf(stderr,
+                "minimal.dtb was built for legacy initramfs boot "
```
shengwen-tw: > minimal.dtb was built for legacy initramfs boot

Is this claim correct?

scripts/build-artifacts.sh
```diff
@@ -0,0 +1,254 @@
...
+                    given, derive ext4.img from it.
+  test-tools        Build test-tools.img. By default this uses the canonical
+                    X11 + DirectFB2 smoke-test recipe.
+  all               Build image, rootfs, and canonical test-tools.
+
+Options:
+  --x11             Select the X11/C++ runtime recipe entry for test-tools. When
+                    any test-tools recipe flag is given, only selected
+                    entries are included.
+  --directfb2-test  Select the DirectFB2 smoke-test recipe entry for test-tools.
```
shengwen-tw: "smoke-test" doesn't sound accurate to me.

main.c
Mes0903: The claim was too broad. `minimal.dtb` is generated from the same DTS and its bootargs depend on the build features, so it is not inherently legacy-only. In an `ENABLE_EXTERNAL_ROOT=0` build, this guard is specifically rejecting a manual launch without `-i rootfs.cpio`.

scripts/prebuilt/artifact-recipes.sh
```diff
@@ -0,0 +1,508 @@
...
+
+# Describe the currently observed lock owner. Waiters compare this identity
+# before removing stale locks so they do not delete a freshly acquired lock.
+buildroot_lock_identity() {
+    local pid_file=$BUILDROOT_LOCK_DIR/pid
+    local pid
+    local inode
+
+    inode=$(ls -id "$BUILDROOT_LOCK_DIR" 2>/dev/null) || inode=missing
+    inode=${inode%% *}
```
visitorckw: Since the OS can recycle pid and reuse inodes, is there a risk that this might lead to subtle race conditions?

visitorckw:
This is a huge.
I'm curious if we might be able to leverage some existing standard gitHub actions like actions/cache or actions/upload-artifact to simplify parts of this?

scripts/prebuilt/artifact-recipes.sh
Mes0903: I checked this again and the concern is valid. This lock is used to serialize source builds that touch the shared `buildroot/` tree. The outer artifact flow may request `Image`, `rootfs.cpio`, and `test-tools.img` separately, but all of those paths can use or mutate Buildroot state such as the checkout, config, host toolchain, and output directories. The lock is there so those sections do not run concurrently while still allowing Buildroot itself to use its own internal parallelism.

The difficult part is automatic stale-lock recovery. The current code uses the lock directory inode plus the recorded PID as a best-effort identity before removing a stale lock. That is not a strict guarantee because both inode numbers and PIDs can be reused. More importantly, with portable shell/filesystem operations we do not have an atomic "remove this lock only if it is still owned by the same observed owner" primitive. A separate read/compare followed by `rm` can still race with another process that removed the stale lock and then created a fresh one.

The stale-lock case we are trying to recover from should only happen when an artifact source build is interrupted after acquiring the lock, for example when the user kills the build, the shell exits unexpectedly, or a CI runner is terminated at the wrong time. That should be much less common than normal lock contention between concurrent artifact builds.

Given that, I plan to remove the best-effort automatic stale-lock deletion. The portable path will use an atomic symlink lock whose target contains an owner token, so acquiring the lock and publishing the owner identity happen in one operation. Normal release will remove the symlink only when the token still matches the current process. If another process observes a stale owner, it will print an actionable diagnostic asking the user/CI job to remove the lock after confirming that no artifact source build is still running, instead of deleting the lock automatically.

This keeps the common path cross-platform and avoids relying on reusable PID/inode identity for correctness.

Mes0903:
Hi @visitorckw,

We already use the standard GitHub Actions primitives for the platform-dependent storage and transport pieces. 'actions/cache' is used for the workflow-local raw artifact cache, and 'actions/upload-artifact' / 'actions/download-artifact' are used for same-workflow handoff from 'guest-artifact-build' to the test and publish jobs.

The custom scripts are deliberately kept as the platform-independent policy layer. They define artifact classes, compute recipe keys, decide whether each class should come from workflow cache, the rolling release, or a source build, validate the release manifest, and stamp raw artifacts with the recipe key that produced them. Those rules should be the same on Linux, macOS, forks, mirrors, and local developer machines.

So the intended boundary is:

- GitHub Actions primitives handle CI storage and transport.
- semu scripts handle artifact identity, freshness, and source-build policy.

That is why the scripts do not try to replace 'actions/cache' or 'actions/upload-artifact'. They sit above those actions so the platform-specific CI mechanics stay in workflow YAML, while the artifact correctness rules remain in repo-owned, testable shell code.

jserv:

